PR #32, opened 2023-02-23
Move code generation code out of src

cbourjau:
This PR moves the opset code generation code out of `src/` to separate it better from the actual source code of the package that is being distributed. I also added a CI step that checks if the code generation runs without errors. It does not check any of the outputs, though. The `pip install pandoc` call in CI pulls in required Python bindings, but not the actual `pandoc` binary.


.github/workflows/ci.yml
```diff
@@ -42,8 +42,38 @@ jobs:
...
           activate-environment: spox
-      - name: Run Unit Tests
+      - name: Run unit tests
         shell: bash -l {0}
         run: |
           pip install .
           pytest
+
+  opset-generation:
+    name: "Linux - opset-generation - Python ${{ matrix.PYTHON_VERSION }}"
```
JakubBachurskiQC, line 52: Nitpick:
```suggestion
    name: "Linux - opset generation - Python ${{ matrix.PYTHON_VERSION }}"
```

.github/workflows/ci.yml
```diff
@@ -42,8 +42,38 @@ jobs:
...
+          miniforge-version: 4.11.0-2
+          use-mamba: true
+          environment-file: environment.yml
+          activate-environment: spox
+      - name: Run code generation
+        shell: bash -l {0}
+        run: |
+          pip install .
+          pip install pandoc
+          python codegen/generate.py
```
JakubBachurskiQC, line 78: Maybe we could add an `.old` suffix to the existing current modules here and then check if any of the files change? We're aiming for reproducibility, after all. Should just require some bash work (I assume the CI succeeds if there's just no non-zero status code?)

codegen/generate.py
```diff
@@ -556,7 +554,9 @@ def main(
         List of template names under ``jinja_templates/extras/`` to add at the end of the code.
         This includes convenience functions that may use the rest of the operator set.
     target
-        Based directory to save the generated operator set file (not counting subdirectory from ``domain``).
+        Based directory to save the generated operator set file (not
+        counting subdirectory from ``domain``). An error is raised if
+        the folder does not exist.
```
JakubBachurskiQC, line 559: Although I'm sure it's based (and the typo is probably quite old 😄), this should probably spell 'base' (plus some other typos):
```suggestion
        Base directory to save the generated operator set file (not
        counting subdirectories from ``domain``). An error is raised if
        it does not exist.
```

docs/manual/overview.rst
```diff
@@ -28,7 +28,8 @@ In Spox, after calling an operator constructor type inference is immediately run
 Spox provides constructor functions for all operators defined by the ONNX specification.
 This includes both the `ai.onnx <https://github.com/onnx/onnx/blob/main/docs/Operators.md>`_ and `ai.onnx.ml <https://github.com/onnx/onnx/blob/main/docs/Operators-ml.md>`_ domain.
 
-Constructors are autogenerated with the ``src/generate.py`` script based on the list of operator schemas provided by ``onnx``. The modules with them may be imported from ``spox.opset.ai.onnx.*``, at a given version of the opset.
+Constructors are autogenerated with the ``codegen/generate.py`` script based on the list of operator schemas provided by ``onnx``.
```
JakubBachurskiQC, line 31: I would opt for a less specific and more commonly seen name, like `tools/`, maybe `tools/codegen/`. Then we'd have `tools/generate.py` & `tools/templates/` (or `tools/codegen_templates/`). `codegen` might evoke strong feelings :)

.github/workflows/ci.yml
cbourjau: Yes, currently it is only the exit code. I am not sure if checking for diffs is more trouble than it is worth. I am afraid that our CI will go red on every `onnx` release because some docstring was updated upstream. I guess we can just try it out and roll it back if need be.

JakubBachurskiQC: As we agreed, let's keep it simple for now and only test for errors.

JakubBachurskiQC (approved):

PR #34, opened 2023-02-24
Add github action to upload to pypi

cbourjau:
I added the relevant token, too.


JakubBachurskiQC (approved):
I don't really know if there's something missing but it seems fine 😉

PR #35, opened 2023-02-27
Run CI on pull requests

xhochy:
`on: push` only works if you push to the repo and not make a pull request from a fork.


cbourjau (approved):
Thanks!

PR #44, opened 2023-03-03
Warn on usage of type membership overload and prepare removal

### LINKED ISSUE #38: Remove operator overload for `Type` membership
Currently, `Type` also exposes a rudimentary 'type membership' semantic via an overload for `<=` with a short documentation. It's is used in some patches and mostly internal code. 
This is a sensitive subject since we have to set a convention of what 'type compatibility' means - I don't recall if it's actually defined in the standard. This feature should instead be in a protected method.

JakubBachurskiQC:
Previously, we exposed `Type.__le__` for checking if a type is essentially a subtype of another type. This was used in some places internally, and was partially documented. This PR moves `Type.__le__` to `Type._subtype(Type) -> bool`, which is a possible (though unlikely) **breaking change**.

Closes #38


cbourjau (approved):

cbourjau (changes requested):
Actually, I am pretty sure that this is not breaking anything, but let's do this by the book with a proper deprecation for at least one version.

cbourjau:
Yes, no need to put it into its own release. Thanks :)

JakubBachurskiQC:
@cbourjau Sorry for the delay - this PR now raises a deprecation warning on usage with a dedicated test, set for `0.6.2`. `0.7.0` can remove the `__le__` method and respective test.

cbourjau (approved):

Issue #25, opened 2023-02-07
Support value promotion (implicit `Constant`)


JakubBachurskiQC:
This is a feature idea that comes up in alternative libraries. It would be useful to be able to express things like `op.add(x, 1)` or even `op.reshape(x, [-1, 2])` without the need for excess baggage like `op.const` or even `op.constant`. 

This raises the following questions:

- opset-independence - should this be done with a `Constant` operator or something else?
- signatures - should we support `npt.ArrayLike` (everything `numpy` takes), `npt.NDArray`/`np.number` (explicit typing only), or maybe just Python types?
- opset generation - can we handle this in a cleaner way than just generating code that constructs the constant?
- typing - in a case like `op.add(x, 1)`, should `1` just become `int64`, or be coerced to whatever `x` is? Should `op.add(x, 1.0)` have `1.0` be `float32` or `float64` (related: Quantco/spox-foss#24). It is technically possible to implement the coercion with schema info in most cases.

I think it's useful to think about these as value promotion could be important for usability long-term.

cbourjau:
I'll close this since this kind of functionality has found its place in ndonnx now, while Spox has settled into being an opinionated yet Pythonic representation of the ONNX standard.

PR #50, opened 2023-03-14
Add unstable `initializer` function

JakubBachurskiQC:
This is a proposal for an opset-independent `initializer` function in `_future`, built using ONNX initialisers.

```py
def initializer(value: npt.ArrayLike, dtype: npt.DTypeLike = None) -> Var:
    ...
```

It follows the same signature as the new version of the extra `ai.onnx::const` (#45), except that it does not depend on `ai.onnx` - and as such does not fix an opset version.

Initialisers have been implemented in Spox for a while but have seen limited usage. Starting to use them is a step towards having a more complete public API & other interesting uses (like #25).

---

Even though this proposal now places `initializer` in `_future`, it remains used in `inline`, as I think it's an appropriate place (it's the first occurrence where we had to construct an opset-independent constant).

Note the branch name is now outdated, as the new function is not stable.


JakubBachurskiQC:
To clarify, on a technical level initialisers seem to have always been about having a global list of constants as you note. Then the ability to make them behave like default model inputs was added, but it remains unsupported by ORT (due to likely a bug).

I feel similarly that there's something off about applying them in terms of output quality. As I mentioned to you, it seems like `onnx.parser`/`onnx.printer` don't actually support initialisers. 

I'm not fully sure whether I'm happy with it becoming entirely public, maybe it should indeed remain named `initializer` and land in `spox._future` until further notice. `op.const` should serve most needs.

JakubBachurskiQC:
@cbourjau This should now be ready for review to be added to `spox._future`. It is applied in `inline` when constant values have to get constructed, replacing the previous mechanism with adding versioned `Constant` operators (in the specific use case of default argument values).

CHANGELOG.rst
```diff
@@ -7,6 +7,21 @@
...
+
+- The opset ``ai.onnx@18`` is now shipped with Spox (version 18 of the default domain, as introduced in ONNX 1.13). To avoid code duplication, unchanged implementations are 'inherited' from the previous version.
+
+**Bug fixes**
+
+- The operator constructor ``cum_sum`` now also has an alias ``cumsum``, to mirror ``numpy``. This alias should be preferred.
+
+**New unstable features**
+
+- ``spox._future.initializer(ArrayLike, DTypeLike = None) -> Var`` is a new function for creating of variables from a constant value. The created value is constructed with ``numpy.array``. This function is opset-independent and is implemented using initializers.
```
cbourjau, line 23:
```suggestion
- ``spox._future.initializer(ArrayLike, DTypeLike = None) -> Var`` is a new function for creating variables from a constant value. The created value is constructed with ``numpy.array``. This function is opset-independent and is implemented using initializers.
```

CHANGELOG.rst
```diff
@@ -7,6 +7,21 @@
 Change log
 ==========
 
+0.7.0 (2023-04-xx)
+------------------
+
+**New features**
```
cbourjau, line 13: Was it deliberate to put all `0.7` changelog entries here?

JakubBachurskiQC: Yes, I forgot to do it on the PR for v18 I had just merged.

PR #51, opened 2023-03-14
Add more guards for `build` args

JakubBachurskiQC:
This is a small improvement that ensures build inputs are actually arguments, instead of other results. This caused a nasty error (`ScopeError`) before. It also checks that the `outputs` are non-empty. Several tests are added to ensure these (and other) invalidations raise.

### DIFF
--- a/src/spox/_public.py
+++ b/src/spox/_public.py
@@ -105,7 +105,8 @@ def build(inputs: Dict[str, Var], outputs: Dict[str, Var]) -> onnx.ModelProto:
         )
     if not all(isinstance(var._op, Argument) for var in inputs.values()):
         raise TypeError(
-            "Build inputs must be argument Vars, and not results of other operations."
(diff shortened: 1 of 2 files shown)


src/spox/_public.py
```diff
@@ -103,6 +103,12 @@ def build(inputs: Dict[str, Var], outputs: Dict[str, Var]) -> onnx.ModelProto:
         raise TypeError(
             f"Build outputs must be Vars, not {set(type(obj) for obj in outputs.values()) - {Var} }."
         )
+    if not all(isinstance(var._op, Argument) for var in inputs.values()):
+        raise TypeError(
+            "Build inputs must be argument Vars, and not results of other operations."
```
cbourjau, line 108:
```suggestion
            "Build inputs must be `Var`s constructed using the `spox.argument` function. "
            "They must not be results of other operations."
```

cbourjau (approved):
Very nice! One minor optional comment.

PR #55, opened 2023-03-17
Fix dependencies with new package versions

JakubBachurskiQC:
Looks like conda has had some package updates which made ORT reveal a (sort of) bad test case in Spox. There's also a problem with pandoc seemingly?

### DIFF
--- a/tests/test_subgraphs.py
+++ b/tests/test_subgraphs.py
@@ -292,7 +292,7 @@ def test_scan_for_product(op, onnx_helper):
 
 def test_sequence_map_for_zip_mul(op, onnx_helper):
     xs, ys = arguments(
-        xs=Sequence(Tensor(numpy.int32)), ys=Sequence(Tensor(numpy.int32))
+        xs=Sequence(Tensor(numpy.int64)), ys=Sequence(Tensor(numpy.int64))
     )
     (zs,) = op.sequence_map(xs, [ys], body=lambda x, y: (op.mul(x, y),))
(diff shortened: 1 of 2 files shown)


JakubBachurskiQC:
Pandoc installed by conda needs to be pinned to `2.19.2` as that's the upper bound from the Python package by the same name (which is in turn at `2.3`). There is a warning now but pandoc `3.1.1` dropped very recently so we didn't upper-bound before. Maybe there is a cleaner solution here, but it would probably involve having the Python package actually on conda.

tests/test_subgraphs.py
```diff
@@ -292,7 +292,7 @@ def test_scan_for_product(op, onnx_helper):
 
 def test_sequence_map_for_zip_mul(op, onnx_helper):
     xs, ys = arguments(
-        xs=Sequence(Tensor(numpy.int32)), ys=Sequence(Tensor(numpy.int32))
+        xs=Sequence(Tensor(numpy.int64)), ys=Sequence(Tensor(numpy.int64))
```
cbourjau, line 295: Interesting! How did that work before?

cbourjau (approved):

tests/test_subgraphs.py
JakubBachurskiQC: Welp, I guess it was just unchecked or there was an implicit cast done. Who knows!
The exact reason in case it's not obvious is that afterwards the arrays are `int64` implicitly.

PR #56, opened 2023-03-17
Fix test warnings caused by pending const change

JakubBachurskiQC:
This PR fixes warnings for `const` in the test suite as introduced in `0.6.1`. It uses a disambiguated form to get a `float32` constant.


cbourjau (approved):
I was about to open an issue about the warnings ❤️

Issue #37, opened 2023-02-27
Style of passing variadic arguments to operator constructors


JakubBachurskiQC:
It's important to resolve this before stabilising. Should variadic arguments:

1. Status quo. Remain passed as `Iterable` (currently `Sequence`, but I think this is unintentional).
2. Instead be passed as a Python variadic `*args`.

```py
# 1.
op.concat([x, y, z], axis=0)
op.loop(M, cond, [a, b, c], body=...)
op.sequence_map(s1, [s2, s3, ...], body=...)

# 2.
op.concat(x, y, z, axis=0)
op.loop(M, cond, a, b, c, body=...)
op.sequence_map(s1, s2, s3, ..., body=...)
```

- Especially the `SequenceMap` example is glaring in (1), as a ≥1 input constraint is enforced by having one normal input and a variadic (≥0) input. 
- However, `Loop` isn't great, as there the separation actually does make sense. Arguably that operator deserves a better interface in general.

I think with a change we might also ensure that we don't get an infinite or otherwise misbehaved iterable as an argument. It's also less awkward to pass a single variadic input. 
This would be a **breaking change** for existing code, but it should be caught by mypy (as in almost all contexts the `Var`s passed are constructed into an iterable at the call-site) and easily fixable by adding the splat operator (`xs -> *xs`).

Related: should we allow subgraph callables to return a single output, implicitly wrapping it in a singleton tuple (`x -> (x,)`)? Think `op.if_(cond, then_branch=lambda: x, else_branch=lambda: y)`

---

A gain to making all arguments `Var | None` is making automatic usage of operator constructors easier. It becomes almost sufficient to do something like 
```py
get_constructor(node.op_type, node.domain)(
  *[get_var(k) for k in node.inputs], 
  **{k: to_py_attr(v) for k, v in node.attributes.items()}
)
```
This is an idea that I've been thinking about for a longer time and I think it could allow for more interesting uses of Spox - for instance in model optimisation or other kinds of in-place modification. The reason this makes sense is because the actual IR treats inputs as a flat list, so introducing more structure might not actually pay off.

The feature I'm talking about here though could still be implemented by parsing the schema for the operator or the signature for the constructor, so this is not a deal-breaker for having it.

JakubBachurskiQC:
[...]

PR #59, opened 2023-03-20
Implementation attempt - hybrid variadic argument passing

JakubBachurskiQC:
This is an attempt (second, actually) at implementing #37. I think this is the simplest implementation I can do and I'm not really happy with it.

A point that was not considered in the proposal directly is that we have been passing variadic arguments as keyword arguments, for instance when doing `op.loop(v_initial=[...], ...)`. This means that the variadic argument cannot be a Python variadic (`*args`), as then it cannot be passed by name. This limits our design space.

Hence, we need to have this sort of signatures (possibly limited by `@typing.overload`, but in general they would be this):

```py
def concat(
    inputs: Union[Var, Iterable[Var]],
    *args: Var,
    axis: int,
) -> Var:
    ...

def loop(
    M: Optional[Var] = None,
    cond: Optional[Var] = None,
    v_initial: Union[Var, Iterable[Var]] = (),
    *args: Var,
    body: Callable[..., Iterable[Var]],
) -> Sequence[Var]:
    ...

def sequence_map(
    input_sequence: Var,
    additional_inputs: Union[Var, Iterable[Var]] = (),
    *args: Var,
    body: Callable[..., Iterable[Var]],
) -> Sequence[Var]:
```

Where the variadic input of a name may stand for either an `Iterable[Var]` of the first of `Var`, the rest of which go into `*args`. This allows passing both in the style of `op.sequence_map(s1, s2, s3, ...)` and `op.concat([x, y, z])`.

However, I'm no longer sure any possible benefits of this are worth the uglier signature. The only use case I had in mind (programmatic constructor calls from ONNX input lists) can just use `inspect` to find there is a `Sequence` argument and the end, and pass to it explicitly.

What are your thoughts on scrapping #37 based on how this attempt turned out, @cbourjau ?

### DIFF
--- a/src/spox/opset/ai/onnx/ml/v3.py
+++ b/src/spox/opset/ai/onnx/ml/v3.py
@@ -916,8 +916,8 @@ def dict_vectorizer(
 
 
 def feature_vectorizer(
-    X: Sequence[Var],
-    *,
+    X: Union[Var, Iterable[Var]],
+    *args: Var,
     inputdimensions: Optional[Iterable[int]] = None,
 ) -> Var:
     r"""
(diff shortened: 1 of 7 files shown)


JakubBachurskiQC:
Closing due to unsatisfactory results along with the source issue #37. Left for future reference.

JakubBachurskiQC:
Closing due to unsatisfactory results along with the source issue #37. Left for future reference.

Issue #58, opened 2023-03-19
Marking deprecated operators in generated opset


JakubBachurskiQC:
ONNX has had 2 deprecated operators - [Scatter](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Scatter) (since 11) and [Upsample](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Upsample) (since 10) - which we generate normally in the opset modules. We should definitely attempt to avoid this and consider strategies for removing them.

**Status quo.** Currently, only Scatter's docstring has a deprecation notice, but Upsample is extremely specific and likely wasn't used by anyone. Such operators might not be implemented by a runtime and pose a possible source of confusing behaviour. 

**Proposed solution.** Raise deprecation warning ASAP, remove completely at `<=1.0`. In the future, deprecated operators shouldn't be inherited on new opset versions.

JakubBachurskiQC:
See discussion in #62 for details on the reasoning for removal of deprecated operators. 

- https://github.com/onnx/sigs/issues/18
- https://github.com/Quantco/spox/pull/62#issuecomment-1478466266

PR #62, opened 2023-03-20
Raise a warning when a deprecated operator is used

JakubBachurskiQC:
This is a first step at #58. Warnings are raised when deprecated operators are used. See the issue for a description.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Change log
 **Pending breaking changes**
 
 - The previously available ``Type <= Type`` (``Type.__le__``) overload is deprecated and will be removed in Spox ``0.7.0``, as it was unintentionally public.
+- Constructors for deprecated ONNX operators (currently ``Scatter`` and ``Upsample``) now raise a warning when they are called. They will be removed entirely in ``0.7.0``.
 
 **Bug fixes**
 
--- a/src/spox/opset/ai/onnx/v17.py
+++ b/src/spox/opset/ai/onnx/v17.py
@@ -13147,7 +13147,8 @@ def scatter(
     """
     warnings.warn(
         "Scatter is a deprecated operator and its constructor should not be used. "
-        "Deprecated constructors will be removed in Spox 0.7.0.",
+    "Building will raise an error. "
+    "Deprecated constructors will be removed in Spox 0.7.0.",
         DeprecationWarning,
         stacklevel=2,
     )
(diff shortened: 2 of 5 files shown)


CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ Change log
 **Pending breaking changes**
 
 - The previously available ``Type <= Type`` (``Type.__le__``) overload is deprecated and will be removed in Spox ``0.7.0``, as it was unintentionally public.
+- Constructors for deprecated ONNX operators (currently ``Scatter`` and ``Upsample``) now raise a warning when they are called. They will be removed entirely in ``0.7.0``.
```
cbourjau, line 16: I don't think we should remove operators from spox which are still in the standard. The warning is super nice for the user though!

cbourjau (changes requested):
I don't think it is spox responsibility to remove the constructors as long as there is an operator of that name and version in the standard. The warning is very nice though. What was your motivation to remove the constructor functions?

JakubBachurskiQC:
No operators were ever removed as far as I'm aware. They only get marked as deprecated from some version onwards and my understanding is that means they are effectively removed. For instance Scatter was renamed in this way to ScatterElements to mirror GatherElements. Definitely feels like we shouldn't just offer them up normally if at some point ONNX (and especially other tools) might start throwing errors about it.

JakubBachurskiQC:
I think this is conclusive to show we shouldn't produce deprecated operators - https://github.com/onnx/sigs/issues/18

JakubBachurskiQC:
Attempting to check a model with a deprecated operator always results in a `ValidationError`, but because in some of the technicalities in how Spox checks individual nodes this only raises when the graph is actually built:

```py
def test_deprecated_raises(op):
    (x,) = arguments(x=Tensor(float, (None,)))
    s = op.const(numpy.array([2.0], numpy.float32))
    with pytest.warns(DeprecationWarning):
        y = op.upsample(x, s)  # Uses the new warning
    graph = results(y=y).with_arguments(x)
    with pytest.raises(Exception):
        graph.to_onnx_model()  # Will always fail since check_model is called
```

cbourjau:
Fair enough. Could you add a sentence to the warning that graphs with deprecated operators fail to build today, please?

cbourjau (approved):

tools/templates/construct.jinja2
```diff
@@ -1,3 +1,10 @@
+{% if schema.deprecated %}
+warnings.warn(
+    "{{ schema.name }} is a deprecated operator and its constructor should not be used. "
+    "Deprecated constructors will be removed in Spox 0.7.0.",
```
JakubBachurskiQC, line 4:
```suggestion
    "Building will raise an error. "
    "Deprecated constructors will be removed in Spox 0.7.0.",
```

src/spox/opset/ai/onnx/v17.py
```diff
@@ -15485,6 +15491,12 @@ def upsample(
     Type constraints:
      - T: `tensor(bool)`, `tensor(complex128)`, `tensor(complex64)`, `tensor(double)`, `tensor(float)`, `tensor(float16)`, `tensor(int16)`, `tensor(int32)`, `tensor(int64)`, `tensor(int8)`, `tensor(string)`, `tensor(uint16)`, `tensor(uint32)`, `tensor(uint64)`, `tensor(uint8)`
     """
+    warnings.warn(
+        "Upsample is a deprecated operator and its constructor should not be used. "
+        "Deprecated constructors will be removed in Spox 0.7.0.",
```
JakubBachurskiQC, line 15496:
```suggestion
    "Building will raise an error. "
    "Deprecated constructors will be removed in Spox 0.7.0.",
```

PR #63, opened 2023-03-20
Remove the opset fixture in the test suite

JakubBachurskiQC:
The `op` fixture is an old antique that has been in the codebase for a long time. It doesn't really make sense to test on multiple opsets, as if the operators don't remain the same, they may have breaking changes. Instead, they should be tested separately - for Spox specifically we only care about checking if we can replicate specific behaviour in the framework, hence we only need to think about fixed versions.
This PR removes the opset fixture. It depends on #57 which should be merged first.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,7 +5,6 @@ import numpy
 import onnxruntime
 import pytest
 
-import spox.opset.ai.onnx.v17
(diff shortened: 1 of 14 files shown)


cbourjau (approved):
Thanks for doing this chore!

PR #74, opened 2023-04-11
Fix issues with non-trivial scope resolution

JakubBachurskiQC:
## Context 

One of the stages of build is resolving where nodes should be put in subgraphs, which is called scope resolution. This creates a scope tree, which describes how scopes are related to each other (a tree is appropriate since with lexical scoping each scope can access values in its ancestors). The algorithm here finds the least enclosing scopes, which attempt to place nodes in the deepest scope possible without invalidating scopes.

## Description

This PR fixes some bugs in scope resolution.

- The implementation of LCA for the scope tree was wrong, which resulted in bugs when the paths to the LCA were not balanced or too close to the root. It has to be implemented in a slightly non-standard way as the tree that is dealt with is dynamic, but it should be correct now.
- The scopes were resolved in an order that was usually topological, but not on more tricky cases. Now it is strictly topological. This ensures nodes are properly initially marked as 'necessarily in this scope' (on that note, I rechecked that LCA seems necessary to satisfy these constraints properly).

There are also some new parameterised tests and a tricky case that I added to stress-test on different kinds of scope trees. 

## Notes

The bug surfaced when implementing a program with complex control flow patterns. It was quite difficult to debug, since:

- It only surfaced as a specific value (an Argument, which was coincidental) not being in scope at the last stage (compilation)
- The traceback is rather inadequate in the situation the build algorithm fails
- The code was quite complex at that point and it was difficult to isolate where the problem even was

The tools useful for fixing this were:

- Minimising the example to only contain the offending loop call
- Introspecting the tracebacks of where specific nodes got constructed (already possible with `Node._traceback`)
- Visualising the graph of operator and subgraph dependencies (under `networkx`)
- Manually naming all the subgraphs in the graph so that looking at relations between them is more useful

Maybe methods for achieving these should be integrated into the codebase to make this sort of problem easier to fix in the future, though the assumption is over time we will encounter them less and less. This has been the first time there were bugs in building - though I fixed two more extra bugs on the way so maybe they're just hard to discover naturally and should be tested more robustly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,12 @@
 Change log
 ==========
 
+0.7.1 (2023-04-xx)
+------------------
+
+- Fix an issue with resolving scopes during the build process. They manifested with a ``KeyError`` in accessing variables in scope on graphs in complex dependencies between subgraphs.
+
+
 0.7.0 (2023-04-04)
 ------------------
 
--- a/src/spox/_build.py
+++ b/src/spox/_build.py
@@ -305,12 +305,13 @@ class Builder:
           graph that we are in (as this is an indirect result to this graph).
           In this case we update the scope to the LCA of this graph's scope and the existing scope.
         - This cannot break the input-scope (1st) constraint (as we only expose more values to scopes),
-          and the 2nd constraint is not moved.
-        - We may recursively descend into subgraphs found by traversing this graph. This will serve to update the scope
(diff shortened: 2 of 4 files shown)


cbourjau (approved):
Thanks for the great summary and the fix of the bugs! I'm not sure what we could do in the future to make this simple. As you say, this also becomes less and less common with every bug found.

PR #76, opened 2023-04-18
Add context managers for global flags

JakubBachurskiQC:
`spox._future` currently has `set_type_warning_level` and `set_value_prop_backend` to be used in edge cases where the defaults don't work. This PR adds context manager versions for these, so that their effects can be better managed.


cbourjau (approved):
Sorry, I forgot about this PR for a bit!

PR #79, opened 2023-05-09
Build graphs with an explicit IR version

cbourjau:
The latest release of `onnx` has introduced a new IR version (9). The function `onnx.helper.make_model` implicitly defaults to the latest function. However, onnxruntime is not yet able to consume models with the latest IR version causing an error. Until the ecosystem had a reasonable amount of time to adopt the new IR version we should not produce models with it.

Setting the IR version like this is a safe way according to an upstream discussion: https://github.com/onnx/onnx/discussions/5203#discussioncomment-5841152

# Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,15 @@
 Change log
 ==========
 
+
+0.7.1 (2023-05-09)
+------------------
+
+**New feature**
+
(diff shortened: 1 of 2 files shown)


cbourjau:
I am not sure what the best procedure would look like, but for the time being this seems to be the simplest place to guarantee that models generated with spox continue to work with onnxruntime using the latest dependencies. If this could be mitigated upstream in the future I'm all for it :)

PR #81, opened 2023-05-18
Raise an error when a subgraph local is leaked to an outer scope

jbachurski:
Though this is quite minor, using side effects in subgraph callbacks is not safe if it causes the 'local' subgraph arguments (as passed to the callback by the operator) to be leaked to an outer scope graph. This obviously can't possibly build (we'd have to move the argument to the outer scope, but this isn't possible). 

I didn't really encounter this issue in the wild except figuring out it is possible in theory. I thought it was worth fixing as it's better to raise in Spox than in ONNX (at which point debugging what happened may be really difficult to someone down the line). 

Now build does an obvious check against this to produce a more helpful error message (a `BuildError` instead of a much delayed ONNX-side `ValidationError`).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,13 @@
 Change log
 ==========
 
+0.8.1 (2023-05-xx)
+------------------
+
+**Bug fixes**
+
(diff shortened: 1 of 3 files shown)


cbourjau (approved):
Thanks for fixing this preemptively! I can imagine that being a nasty thing to debug downstream!

PR #83, opened 2023-06-10
Generate opset 19 & fix Split 18 constructor

### LINKED ISSUE #80: Signature of `spox.opset.ai.onnx.v18.split` is wrong
Opset 18 added an explicit attribute for the number of (variadic) outputs in Split. Unfortunately, the way this is generated in Spox isn’t particularly robust and currently the signature contains both the new attribute and the old workaround for specifying this (only the latter of which works, but IIRC both are required).

https://github.com/Quantco/spox/blob/5e83a185a5e2868d560cb888bd9831689ee3fc01/src/spox/opset/ai/onnx/v18.py#L2713-L2720

Compare to opset 17, which used just the workaround:

https://github.com/Quantco/spox/blob/5e83a185a5e2868d560cb888bd9831689ee3fc01/src/spox/opset/ai/onnx/v17.py#L14302-L14308

Bug workaround: use split from v17 and let the adapter handle this.

jbachurski:
- Closes #80 
- Generates `ai.onnx@19`, bumping to **ONNX 1.14**
- Removes the dependency on the `pandoc` _Python package_, as it is incompatible with Pandoc 3.*, which is the current version shipped on modern distributions. We use a very small subset of Pandoc, so using `subprocess` directly is okay.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,18 @@
 Change log
 ==========
 
+0.9.0 (2023-06-xx)
+------------------
+
+**New features**
+
+- The opset ``ai.onnx@18`` (ONNX 1.14) is now shipped with Spox.
+
+**Bug fixes**
+
+- The constructor for ``ai.onnx@18::Split`` is no longer generated incorrectly. No extraneous attribute is generated anymore, and the ``num_outputs`` attribute is marked as required (so that Spox can infer the number of outputs).
+
+
 0.8.1 (2023-05-xx)
 ------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ Change log
 
 **New features**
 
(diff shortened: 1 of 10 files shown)


jbachurski:
Apparently ONNX 1.14 is broken in conda, but it would be nice to have it running in CI now?

CHANGELOG.rst
```diff
@@ -7,6 +7,18 @@
 Change log
 ==========
 
+0.9.0 (2023-06-xx)
+------------------
+
+**New features**
+
+- The opset ``ai.onnx@18`` (ONNX 1.14) is now shipped with Spox.
```
```suggestion
- The opset ``ai.onnx@19`` (ONNX 1.14) is now shipped with Spox.
```

cbourjau (approved):
Thanks so much for this PR! 

It seems there was a timeout while uploading some of the conda packages of `onnx`. I just re-ran the build step which I hope fixes the issue. If not, I'll look into it tomorrow!

The PR looks great to me but lets fix the CI before merging.

Issue #82, opened 2023-05-18
Tracking issue for full-feature inline


jbachurski:
`inline` is currently missing:

- [ ] Support for subgraphs (they should get prefixed properly, which ONNX renaming doesn't do)
- [ ] Support for functions (such local/dependency functions should be propagated to the built model)
- [ ] Capability not do add extraneous Identity nodes for all model input/outputs (requires better renaming capabilties)

Most of these can be fixed by reimplementing renaming values in graphs, as the ONNX feature is rather lacking.

PR #84, opened 2023-06-10
Implement a custom graph renaming routine for inline

jbachurski:
Addresses #82 by avoiding adding extraneous Identity nodes when inlining and supporting subgraph renaming

- Reimplements the `onnx.compose.add_prefix_graph` to make it more flexible.
- Slightly expands Spox scopes to support 'anonymous' reserved names, which don't have an associated `Var`/`Node`, but are nevertheless marked as taken. This is because not all inlined nodes/values have a respective `Var`/`Node`.

### DIFF
--- a/src/spox/_inline.py
+++ b/src/spox/_inline.py
@@ -35,7 +35,15 @@ def rename_in_graph(
         if nd.name and rename_node is not None:
             nd.name = rename_node(nd.name)
         if rename_op is not None:
-            nd.domain, nd.op_type = rename_op(nd.domain, nd.op_type)
+            # This is a bit elaborate, but we do it this way as
+            # an unset domain field is different from an empty one.
+            if nd.HasField("domain"):
+                nd.domain, nd.op_type = rename_op(nd.domain, nd.op_type)
+            else:
+                # An empty domain is the default domain (ai.onnx)
+                domain, nd.op_type = rename_op("", nd.op_type)
+                if domain:  # Only set the domain explicitly if it's changing
+                    nd.domain = domain
         for seq in (nd.input, nd.output):
             for i, name in enumerate(seq):
                 seq[i] = rename(name)
--- a/src/spox/_inline.py
+++ b/src/spox/_inline.py
@@ -50,6 +50,18 @@ def rename_in_graph(
                         rename_op=rename_op,
                     )
                 )
+            elif isinstance(attr, list) and all(
+                isinstance(g, onnx.GraphProto) for g in attr
+            ):
+                for i in range(len(attr)):
+                    attr_proto.graphs[i].CopyFrom(
+                        rename_in_graph(
+                            attr[i],
+                            rename,
+                            rename_node=rename_node,
+                            rename_op=rename_op,
+                        )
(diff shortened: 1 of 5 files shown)


src/spox/_inline.py
```diff
@@ -13,6 +15,48 @@
...
+    graph.CopyFrom(graph_)
+
+    for p in itertools.chain(graph.input, graph.initializer):
+        p.name = rename(p.name)
+    for si in graph.sparse_initializer:
+        si.values.name = rename(si.values.name)
+        si.indices.name = rename(si.indices.name)
+
+    for nd in graph.node:
+        if nd.name and rename_node is not None:
```
cbourjau, line 35: Are nodes really allowed to have a missing `name` / `op_type`?

src/spox/_public.py
```diff
@@ -218,13 +218,11 @@ def inline(model: onnx.ModelProto) -> _InlineCall:
     _signature_msg = f"signature {in_names}{_defaults_msg} -> {out_names}"
 
     model = _copy_model(model)
-    # FIXME: Renaming does not work on subgraphs as of ONNX 1.13/1.14.
     for node in model.graph.node:
         for attr in node.attribute:
-            if attr.HasField("g") or attr.graphs:
+            if attr.graphs:
```
cbourjau, line 223: Is there any technical reason why we cannot support this now?

cbourjau (approved):
Great! Thanks! I just have some minor suggestions/questions.

> This is because not all inlined nodes/values have a respective Var/Node.

Could you expand on this a bit? Do you mean the nodes inside a subgraph?

src/spox/_inline.py
jbachurski: https://github.com/onnx/onnx/blob/main/docs/IR.md#nodes

> name | string | An optional name of the node, used for diagnostic purposes only.

I think it's often left empty/unset in practice, though Spox does build the nodes with its internal naming.

You're right about `op_type`, though. FYI, replacing `op_type` is there for future-proofing as inlining models with functions might require renaming some local functions.

jbachurski:
Looks like due to the conda ONNX version bump, the CI is now failing on opset generation, so #83 will fix this :)

jbachurski:
I also added an explicit proto-based test, and one for subgraph lists as there is no practical example that can be constructed there (no node uses a list of subgraphs...). This is since we are now straying from the ONNX utility which would have some tests itself. Maybe I'll upstream the new one at some point?

> > This is because not all inlined nodes/values have a respective Var/Node.
> 
> Could you expand on this a bit? Do you mean the nodes inside a subgraph?

@cbourjau Yeah - say an inlined `ModelProto` has an input X and output Y, and an intermediate Z. Then there is a `Var` corresponding to X and another for Y, but there will be no explicit variable for Z in the Spox graph. However, it does have to reserve some name for itself in-scope to avoid e.g. other inlines of the same model using that name. So the `ScopeSpace` object marks whatever name was used for Z as taken with the newly applied patch.

cbourjau:
Thanks for the clarification :)

PR #85, opened 2023-06-10
Allow creation of untyped variables

jbachurski:
This is a simple proposal to allow `argument(None)`, i.e. untyped arguments and hence variables. Such variables would have no type checking done.

The goal of this is to allow defining functions by defining a model like usual, but with untyped variables. This is because functions are not typed in ONNX at all. 

I think this is a start to getting rid of the whole facility related to `_function.Function` - as this is really messy when it interferes with the 'real' Spox graph - and move everything to a new `build`-like function which would construct a `FunctionProto`. 

The last two steps to refactor this would be to make sure attribute references work (`_attr.Ref`) and making `Function` just a thin wrapper around `FunctionProto` - or possibly remove `Function` entirely and piggyback off of `Inline` which could then support functions itself. 

For example, the goal would be to get something like:
```py
t = attribute_ref(int)  # Make a new attribute reference for this function
x = argument(None)  # Untyped arguments
y = argument(None)
z = op.add(x, y)  # Usual graph operators
w = op.mul(z, op.constant(value_int=t))  # Using the function attribute here
# List and name all attributes for this function
function: onnx.FunctionProto = build_function({"x": x, "y": y}, {"w": w}, {"t": t})  
# This might actually also need to return a list of FunctionProto objects, 
#  since functions may depend on each other and we have to manage those dependencies...

...

a = argument(Tensor(int, ()))
# inline_function - like normal inline, but takes a FunctionProto
#  (could be a thin wrapper on inline with support for functions)
#  type inference & checking can be implemented here by expanding the function body
(b,) = inline_function(function)(a, a)
# build collects the function_proto used in the graph and places it in the model
# manages any colliding function names by adding a discriminator
model = build({"a": a}, {"b": b})  
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@
 Change log
 ==========
 
+0.9.0 (2023-06-xx)
+------------------
+
+**New features**
+
+- ``argument`` can now take ``None``, indicating the argument variable is untyped.
+
+
 0.8.1 (2023-05-xx)
 ------------------
 
--- a/src/spox/_internal_op.py
+++ b/src/spox/_internal_op.py
@@ -64,7 +64,7 @@ class Argument(_InternalNode):
 
     @dataclass
     class Attributes(BaseAttributes):
-        type: AttrType
+        type: Optional[AttrType] = None
         name: Optional[AttrString] = None
         default: Optional[AttrTensor] = None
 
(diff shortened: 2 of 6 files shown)


jbachurski:
I’m closing this as there hasn’t been a need for implementing functions in Spox. They can be hacked in with inlining and adding them to the model manually which shouldn’t be that big of a hassle.

PR #87, opened 2023-06-16
Fix dtype aliasing removal

AtanasDimitrovQC:
Currently because of numpy type-aliasing there is type-inference error in this code:
```
op.const(np.iinfo(np.int64).max + 1)
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Change log
 **Bug fixes**
 
 - The constructor for ``ai.onnx@18::Split`` is no longer generated incorrectly. No extraneous attribute is generated anymore, and the ``num_outputs`` attribute is marked as required (so that Spox can infer the number of outputs).
+- Fix an issue where op.const handled numbers in the range(INT64MAX, UINT64MAX) incorrectly
 
 **Other changes**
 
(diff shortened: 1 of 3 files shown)


jbachurski:
Hi! Thank you for the PR and catching this.

This seems to be related to a problem that I had found also interfacing with ORT before, and which only had a fix in `_utils.dtype_to_tensor_type`. 

I'll pick up your PR as I also wanted to add a test and better document the change.

CHANGELOG.rst
```diff
@@ -56,7 +57,7 @@ This version is intended as a release candidate for ``1.0.0``.
 
 **Breaking changes**
 
-- The typing rules of the (previously partially documented) extra constructor ``const`` have changed. Its signature is now ``const(npt.ArrayLike, npt.DTypeLike = None) -> Var``. In particular, ``const`` of a Python ``float`` no longer becomes ``float32``, but ``float64`` like numpy - this is a **breaking change**. The operator is redefined to be equivalent to ``constant(numpy.array(value), dtype)``, instead of a complex set of cases like before. 
+- The typing rules of the (previously partially documented) extra constructor ``const`` have changed. Its signature is now ``const(npt.ArrayLike, npt.DTypeLike = None) -> Var``. In particular, ``const`` of a Python ``float`` no longer becomes ``float32``, but ``float64`` like numpy - this is a **breaking change**. The operator is redefined to be equivalent to ``constant(numpy.array(value), dtype)``, instead of a complex set of cases like before.
```
jbachurski, line 60: There's no problem here, except it does pollute the diff, unless you really want to remove the space :)

tests/test_value_propagation.py
```diff
@@ -151,3 +151,10 @@ def test_give_up_silently():
 
 def test_non_ascii_characters_in_string_tensor():
     op.cast(op.constant(value_string="FööBär"), to=str)
+
+
+def test_propagated_value_does_not_alias_dtype():
+    # Ensures that platform-dependent dtypes aren't accidentally propagated
+    x = numpy.iinfo(numpy.int64).max + 1
+    # Without the explicit astype(uint64), x actually ends up being ulonglong
+    assert_equal_value(op.const(x), numpy.array(x).astype(numpy.uint64))
```
jbachurski, line 160: Please do add tests like this to your PRs in the future :wink: 
Just pick the test file which matches the tested module the most.

jbachurski (approved):
Just a few informative comments.
I'll clarify that the actual failing module here wasn't type checking (type inference), but value propagation. It was devised as a separate module as it brings its own host of issues and is largely separate from type inference (it is also effectively still experimental). 
As for the source, value propagation sanity-`.check()`s whether the propagated value (likely sourced from some runtime - which we don't really have control over) actually matches the type provided by type inference. In this case, this mechanism failed and raised a warning.

CHANGELOG.rst
AtanasDimitrovQC: My auto-whitespace-removal was on and decided to leave it in the pr :^).

PR #88, opened 2023-06-16
Remove outdated patches for upstream libraries

jbachurski:
No changes here, just removing old code which is no longer necessary.

There are still a few patches around that are no longer necessary as they have been fixed upstream.

- https://github.com/onnx/onnxmltools/issues/575
- Reference implementation: Incosistent handling of optionals (`OptionalHasElement`), missing string cast support (`Cast`, `to=str`)

Related: should we pin ONNX releases with Spox, since we don't run tests on them? (to be clear, they would almost definitely fail - and e.g. this patch assumes that `onnx>=1.14`, while it seems that ONNX releases usually also bring breaking changes)


cbourjau (approved):
Thanks for the PR!

I would hope that after [pinning the IR version](https://github.com/Quantco/spox/blob/2607981b68cfaa022e447a50b1f9d3614f1026e0/src/spox/_graph.py#L357) we should be fine with respect to breaking changes. Pinning dependencies has a tendency to always turn into trouble down the road.

PR #92, opened 2023-06-27
Fix value-propagation for subgraphs

AtanasDimitrovQC:
Change the way value-propagation works in regards to operators with subgraphs. 

This was currently missing as value-propagation on subgraphs seems to be slow, however as it is an optional feature in my opinion we should strive for completeness instead of performance.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,10 +7,13 @@ import pytest
 
 from spox import _value_prop
 from spox._debug import show_construction_tracebacks
-from spox._future import set_type_warning_level
+from spox._future import (
+    ValuePropBackend,
+    set_type_warning_level,
+    set_value_prop_backend,
+)
 from spox._graph import Graph
 from spox._node import TypeWarningLevel
-from spox._future import set_value_prop_backend, ValuePropBackend
 
 set_value_prop_backend(ValuePropBackend.ONNXRUNTIME)
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,9 @@ from spox._debug import show_construction_tracebacks
 from spox._future import set_type_warning_level
 from spox._graph import Graph
 from spox._node import TypeWarningLevel
+from spox._future import set_value_prop_backend, ValuePropBackend
+
+set_value_prop_backend(ValuePropBackend.ONNXRUNTIME)
 
 set_type_warning_level(TypeWarningLevel.CRITICAL)
 _value_prop.VALUE_PROP_STRICT_CHECK = True
--- a/tests/test_subgraphs.py
+++ b/tests/test_subgraphs.py
@@ -440,9 +440,8 @@ def test_subgraph_argument_leak_caught():
         ii = i
         return [op.const(True), i]
 
-    (m,) = arguments(m=Tensor(int, ()))
-    (r,) = op.loop(M=m, v_initial=[], body=fun)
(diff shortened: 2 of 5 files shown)


tests/conftest.py
AtanasDimitrovQC, line 1: The reference implementation currently has problems with `test_loop`.

tests/test_subgraphs.py
```diff
@@ -440,9 +440,8 @@ def fun(i, c):
         ii = i
         return [op.const(True), i]
 
-    (m,) = arguments(m=Tensor(int, ()))
-    (r,) = op.loop(M=m, v_initial=[], body=fun)
-    graph = results(_=op.add(r, ii))
-
     with pytest.raises(BuildError):
+        (m,) = arguments(m=Tensor(int, ()))
```
AtanasDimitrovQC, line 444: The error is now caught earlier, since it happens in the value propagation.

cbourjau:
We just had a discussion about this in person. The certain take-away is that experimental features should not deteriorate the user experience for stable features. Therefore, this current approach is unfortunately not doable.

Below is a non-exhaustive list of options to go forward to get value propagation for nodes with subgraphs:
- Make this implementation opt-in
- Use the fact that subgraphs with no inputs produce constants as far as value propagation is concerned
- Keep track of the scope in which a node was created. This would enable us to:
  - Cheaply find sub-graph boundaries which work for the value propagation (unless the user creatively used side effects)
  - Fail the value propagation if the user did go crazy with side effects
  - Use this information to warn the user when `spox.build` is called that they used a node somewhere that was created in a "sibling" scope (e.g. a node created in an `If`-node's `then_branch` was subsequently used in the `else_branch`)

cbourjau (commented):
We just had a discussion about this in person. The certain take-away is that experimental features should not deteriorate the user experience for stable features. Therefore, this current approach is unfortunately not doable.

Below is a non-exhaustive list of options to go forward to get value propagation for nodes with subgraphs:
- Make this implementation opt-in
- Use the fact that subgraphs with no inputs produce constants as far as value propagation is concerned
- Keep track of the scope in which a node was created. This would enable us to:
  - Cheaply find sub-graph boundaries which work for the value propagation (unless the user creatively used side effects)
  - Fail the value propagation if the user did go crazy with side effects
  - Use this information to warn the user when `spox.build` is called that they used a node somewhere that was created in a "sibling" scope (e.g. a node created in an `If`-node's `then_branch` was subsequently used in the `else_branch`)

jbachurski:
I do agree value propagation should remain opt-in except when of trivial cost (and in trivial cases - like currently in expressions which are constant anyway). 
@AtanasDimitrovQC - So we can't always attempt to build the subgraph to test whether it can have values propagated as this will cause operator constructors to no longer be constant time.

I do remember your idea, @cbourjau, of trying to recover the scope information from context. I'm still not convinced it would be painless, however. On one hand it might make things more intuitive, but on the other it does introduce global side effects which might interact in unexpected ways. However, I'm not sure if the proposal is necessary here - and I think it would be appropriate for it to be considered along with a larger build refactor.

I'll take another look at the PR later when I have time, but as it stands do reconsider its necessity. Value propagation was a simple feature to improve some testing experience. Maybe it would be sufficient to compute and implement the `Var._value` in downstream code with some custom mechanism tailored to the use case?

cbourjau:
I'll close this for now since any solution does require a redesign of this PR anyhow.

PR #97, opened 2023-07-28
Fix bug where node adaption failed for nodes with repeated inputs

cbourjau:
Node-adaption currently fails if the adapted node has repeating inputs (e.g. the result of `op.equal(a, a)`). This PR addresses that issue.

# Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@
 Change log
 ==========
 
+0.9.1 (2023-xx-xx)
+------------------
+
+**Bug fix**
+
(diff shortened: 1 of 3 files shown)


jbachurski (approved):
LGTM. 
Feel free to have a second review.
This logic has always been tricky and refactoring it might be worth it at some point 😃

cbourjau:
Thanks for the quick review! I ran into another similar issue where we need to adapt nodes in subgraphs. I'll make a separate PR for that next week. This code path is now getting some real-world use since the release of opset version 19.

PR #105, opened 2023-09-15
Forego version adaption of inlined models if no nodes are from the default domain

cbourjau:
Calling `onnx.version_converter.convert_version` with a model that has no nodes from the default domain makes it fail. We should avoid calling it in such a case.

### DIFF
--- a/src/spox/_adapt.py
+++ b/src/spox/_adapt.py
@@ -73,6 +73,10 @@ def adapt_inline(
     source_version = max({v for d, v in node.opset_req if d in ("", "ai.onnx")})
     target_version = target_opsets[""]
 
+    # convert_version fails if the inlined model does not import the default domain
+    seen_domains = {prot.domain for prot in protos}
+    if not seen_domains & {"", "ai.onnx"}:
+        return protos
     if source_version != target_version:
         target_model = onnx.version_converter.convert_version(
             node.model, target_version
--- a/tests/test_adapt.py
+++ b/tests/test_adapt.py
@@ -188,3 +188,48 @@ def test_inline_model_custom_node_only():
     # Add another node to the model to trigger the adaption logic
     c = op18.identity(b)
     build({"a": a}, {"c": c})
+
+
+@pytest.mark.skip(
+    reason="Adapting custom nodes (including their subgraphs) is currently not supported"
+)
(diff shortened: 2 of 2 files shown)


adityagoel4512:
I'll work on a patch this for "ai.onnx.ml" upstream.

src/spox/_adapt.py
```diff
@@ -73,6 +73,10 @@ def adapt_inline(
     source_version = max({v for d, v in node.opset_req if d in ("", "ai.onnx")})
     target_version = target_opsets[""]
 
+    # convert_version fails if the inlined model does not import the default domain
+    seen_domains = {prot.domain for prot in protos}
+    if not seen_domains & {"", "ai.onnx"}:
```
jbachurski, line 78: We should probably make these sort of checks consistent across the codebase.

jbachurski (approved):
LGTM

src/spox/_adapt.py
cbourjau: On the other hand: In this case, we are only interested in the domains used in this particular subgraph. It seems rather fitting to check that here rather than elsewhere. It is also quite specific. We check the domain of the nodes in `protos` rather than those in `node.model` directly.

cbourjau:
While typing the other comment I also wondered about custom operators with GraphProto attributes which may have standard nodes in them. I'll add a test on Monday.

src/spox/_adapt.py
jbachurski: I guess there two main things to consider here:

1. The one I had in mind was checking for the default domain, which is often checked by both `""` and `"ai.onnx"`.

2. On the other, there's finding all the opsets in a model (or node). As you noted, there's the subgraphs which I forgot about.

I think either could be more standardised, but whatever works, probably just works.

cbourjau: I agree. There should be an obvious place where we get that kind of information. For now I think we may file this under "refactoring the build step" Todo-item.

cbourjau:
About the nested subgraphs. I added a `XFail` test case for future reference. In the end, is a special case of a general feature: Being able to conveniently customize the node-adaption logic for some nodes/domains. This is a feature we will want in the future, but I think we would need some larger refactoring that is out-of-scope of this PR.

Issue #103, opened 2023-09-06
Inline call errors if value propagation encounters unknown operators


JoelHuberQC:
## Description
When passing a model to inline which contains an unknown operator, the onnx reference value propagation backend returns an empty dictionary (see _value_prop:160). However, this makes inline error out, since inline expects to receive values from the value propagation backend. In particular, the error we get is some random KeyError when it is trying to extract the values.

I would expect that in this case, value propagation just silently fails, without crashing the code.

## How to Reproduce:
```python
import numpy as np
from onnx import helper
from onnx.checker import check_model
from onnxruntime import InferenceSession
from spox import inline
from spox.opset.ai.onnx import v18 as op

model_input = [helper.make_tensor_value_info("X", elem_type=8, shape=("X",))]
model_output = [
    helper.make_tensor_value_info("y", elem_type=8, shape=("y", "max_words"))
]

# Take the tokenizer from the microsoft contrib domain for example
nodes = [
    helper.make_node(
        "Tokenizer",
        inputs=["X"],
        outputs=["y"],
        domain="com.microsoft",
        mincharnum=1,
        mark=0,
        pad_value="",
        separators=[" "],
    )
]

graph = helper.make_graph(
    nodes,
    "Tokenizer",
    model_input,
    model_output,
)

model = helper.make_model(
    graph,
    opset_imports=[
        helper.make_opsetid("", 18),
        helper.make_opsetid("com.microsoft", 1),
    ],
)

# Model is valid and works on onnxruntime
check_model(model, full_check=True)
ort_output = InferenceSession(model.SerializeToString()).run(
    None, {"X": np.array(["Test Test"])}
)
print(ort_output)

# This crashes since the value prop backend doesn't give an output
spox_const_res = inline(model)(X=op.const(["Test Test"], dtype=np.str_))
```

## Expected Behaviour
Code works, no error, but no value propagation is performed.

## Actual Behaviour
The script throws an error:
```
Traceback (most recent call last):
 [...]
  File "/path/to/spox/_inline.py", line 138, in propagate_values
    return {
  File "/path/to/spox/_inline.py", line 139, in <dictcomp>
    f"outputs_{k}": unwrap_feed(var.unwrap_type(), output_feed[o.name]).value
KeyError: 'y'
```

jbachurski:
[...]

PR #114, opened 2023-10-23
Only propagate produced values

adityagoel4512:
Closes https://github.com/Quantco/spox/issues/103. Given value propagation is best effort anyway it seems safe to fish out the propagated values on a best effort basis, but let me know if I'm not considering something here.

### DIFF
--- a/src/spox/_value_prop.py
+++ b/src/spox/_value_prop.py
@@ -169,9 +169,12 @@ def _run_onnxruntime(
     # Silence possible warnings during execution (especially constant folding)
     options = onnxruntime.SessionOptions()
     options.log_severity_level = 3
-    session = onnxruntime.InferenceSession(model.SerializeToString(), options)
-    output_names = [output.name for output in session.get_outputs()]
-    output_feed = dict(zip(output_names, session.run(None, input_feed)))
+    try:
+        session = onnxruntime.InferenceSession(model.SerializeToString(), options)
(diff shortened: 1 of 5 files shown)


cbourjau (changes requested):
This PR is definitely the correct way! Currently, there is [error handling](https://github.com/Quantco/spox/blob/main/src/spox/_standard.py#L181-L186) for "standard nodes". We can probably remove that code now. It would still be nice to raise a debug warning of that kind in the new implementation, though.

src/spox/_value_prop.py
```diff
@@ -169,9 +174,16 @@ def _run_onnxruntime(
...
-    session = onnxruntime.InferenceSession(model.SerializeToString(), options)
-    output_names = [output.name for output in session.get_outputs()]
-    output_feed = dict(zip(output_names, session.run(None, input_feed)))
+    try:
+        session = onnxruntime.InferenceSession(model.SerializeToString(), options)
+        output_names = [output.name for output in session.get_outputs()]
+        output_feed = dict(zip(output_names, session.run(None, input_feed)))
+    except Exception as e:
+        logging.debug(
+            f"Value propagation in {model} on the ONNX reference implementation failed with - "
```
cbourjau, line 183:
```suggestion
            f"Value propagation in {model} on the onnxruntime failed with - "
```

cbourjau (approved):
Minor copy-pasta mistake. Otherwise, good to go!

PR #122, opened 2023-11-28
More fine-grained validation

SimeonStoykovQC:
Please share your thoughts on this. It's solving a few problems:
- ~30% speedup when instantiating (noticeable in operators with big attributes)~ - we would like to try end-to-end caching first, which should give us a bigger speedup
- Addressing https://github.com/Quantco/spox/pull/119#issuecomment-1827399762 : consistent exception source when spox attribute type checks fail. ~Note we had to omit the call to `.encode()` in the string-like attributes ([ONNX's make_attribute takes care of that](https://github.com/onnx/onnx/blob/5be7f3164ba0b2c323813264ceb0ae7e929d2350/onnx/helper.py#L906)).~ We are now [catching this error](https://github.com/Quantco/spox/blob/4918f3e303d1ca7dc5d87fdbf53b7449fac95d39/src/spox/_attributes.py#L228-L236)
- Custom exception

## TODOs

1. [x] TODO in code: make sure we never use the actual attribute proto, just its type, in the `_Ref` class
2. [x] TODO in code: is there a more elegant way to implement `_deduce_type`, and is the copy-paste worth it
3. [x] Added a `CHANGELOG.rst` entry

## ~Speedup~ (not included in this PR in favour of future caching efforts)
Before:
image

Now:
image

### DIFF
--- a/src/spox/_attributes.py
+++ b/src/spox/_attributes.py
@@ -2,17 +2,7 @@ import abc
 import collections.abc
 import numbers
 from abc import ABC
-from typing import (
-    Any,
-    ClassVar,
-    Generic,
-    Iterable,
-    Optional,
-    Tuple,
-    Type,
-    TypeVar,
-    Union,
-)
+from typing import Any, Generic, Iterable, Optional, Tuple, Type, TypeVar, Union
 
 import numpy as np
 import numpy.typing as npt
@@ -41,7 +31,6 @@ class AttributeTypeError(TypeError):
 
 class Attr(ABC, Generic[T]):
     _value: Union[T, "_Ref[T]"]
-    _attribute_proto_type_int: ClassVar[int]
 
     def __init__(self, value: Union[T, "_Ref[T]"]):
         self._value = value
@@ -60,8 +49,13 @@ class Attr(ABC, Generic[T]):
 
     def _validate(self):
         if self._to_onnx_type_deref() != self._attribute_proto_type_int:
+            if isinstance(self.value, tuple) and len(self.value):
+                tuple_types = ", ".join({type(v).__name__ for v in self.value})
+                value_type = f"tuple[{tuple_types}, ...]"
+            else:
+                value_type = type(self.value).__name__
             raise AttributeTypeError(
-                f"Unable to instantiate `{type(self).__name__}` with value of type `{type(self.value).__name__}`."
+                f"Unable to instantiate `{type(self).__name__}` with value of type `{value_type}`."
             )
 
     def _to_onnx(self, key: str) -> AttributeProto:
@@ -75,9 +69,14 @@ class Attr(ABC, Generic[T]):
         return self._to_onnx_type_deref()
 
     def _to_onnx_type_deref(self) -> AttributeProto.AttributeType:
-        """Conversion method that only computes the type for the dereferenced case. It should match the type produced
-        by _to_onnx_deref"""
+        """Conversion method that only computes the type for the dereferenced case. It must match the type produced
+        by _to_onnx_deref, but could be more efficient."""
         return self._to_onnx_deref("dummy").type
+
+    @property
+    @abc.abstractmethod
+    def _attribute_proto_type_int(self) -> int:
+        raise NotImplementedError
 
     @abc.abstractmethod
     def _to_onnx_deref(self, key: str) -> AttributeProto:
@@ -132,7 +131,6 @@ class AttrString(Attr[str]):
     _attribute_proto_type_int = AttributeProto.STRING
 
     def _to_onnx_deref(self, key: str) -> AttributeProto:
-        # Strings are bytes on the onnx side
         return make_attribute(key, self.value)
 
 
@@ -207,60 +205,74 @@ class _AttrIterable(Attr[Tuple[S, ...]], ABC):
     ) -> Optional[AttrIterableT]:
         return cls(tuple(value)) if value is not None else None
 
+    def _to_onnx_deref(self, key: str) -> AttributeProto:
+        return make_attribute(key, self.value, attr_type=self._attribute_proto_type_int)
+
+    def _to_onnx_type_deref(self) -> AttributeProto.AttributeType:
+        return _deduce_type(self.value)
+
 
 class AttrFloat32s(_AttrIterable[float]):
     _attribute_proto_type_int = AttributeProto.FLOATS
 
     def _to_onnx_deref(self, key: str) -> AttributeProto:
-        # ensure values are all floats
-        values = [float(v) for v in self.value]
-        return make_attribute(key, values, attr_type=AttributeProto.FLOATS)
+        return make_attribute(
+            key, self._transformed(), attr_type=self._attribute_proto_type_int
+        )
 
     def _to_onnx_type_deref(self) -> AttributeProto.AttributeType:
-        values = [float(v) for v in self.value]
-        return _deduce_type(values)
+        return _deduce_type(self._transformed())
+
+    def _transformed(self) -> Tuple[float, ...]:
+        try:
+            return tuple(float(v) for v in self.value)
+        except ValueError as e:
+            raise AttributeTypeError("Attribute values don't seem to be floats.") from e
 
 
 class AttrInt64s(_AttrIterable[int]):
     _attribute_proto_type_int = AttributeProto.INTS
-
-    def _to_onnx_deref(self, key: str) -> AttributeProto:
-        return make_attribute(key, self.value, attr_type=AttributeProto.INTS)
-
-    def _to_onnx_type_deref(self) -> AttributeProto.AttributeType:
-        return _deduce_type(self.value)
 
 
 class AttrStrings(_AttrIterable[str]):
     _attribute_proto_type_int = AttributeProto.STRINGS
 
     def _to_onnx_deref(self, key: str) -> AttributeProto:
-        return make_attribute(key, self.value, attr_type=AttributeProto.STRINGS)
+        return make_attribute(
+            key, self._transformed(), attr_type=self._attribute_proto_type_int
+        )
 
     def _to_onnx_type_deref(self) -> AttributeProto.AttributeType:
-        return _deduce_type(self.value)
+        return _deduce_type(self._transformed())
+
+    def _transformed(self) -> Tuple[bytes, ...]:
+        try:
+            return tuple(v.encode() for v in self.value)
+        except AttributeError as e:
+            raise AttributeTypeError(
+                "Attribute values don't seem to be strings."
+            ) from e
 
 
 class AttrTensors(_AttrIterable[np.ndarray]):
     _attribute_proto_type_int = AttributeProto.TENSORS
 
     def _to_onnx_deref(self, key: str) -> AttributeProto:
-        tensors = [from_array(t) for t in self.value]
-        return make_attribute(key, tensors, attr_type=AttributeProto.TENSORS)
+        return make_attribute(
+            key, self._transformed(), attr_type=self._attribute_proto_type_int
+        )
 
     def _to_onnx_type_deref(self) -> AttributeProto.AttributeType:
-        tensors = [from_array(t) for t in self.value]
-        return _deduce_type(tensors)
+        return _deduce_type(self._transformed())
+
+    def _transformed(self) -> Tuple[TensorProto, ...]:
+        return tuple(from_array(v) for v in self.value)
 
 
 def _deref(ref: _Ref[T]) -> T:
     if isinstance(ref._concrete._value, _Ref):
         return _deref(ref._concrete._value)
     return ref._concrete._value
-
-
-def _make_attribute_type(value: Any) -> AttributeProto:
-    return AttributeProto(type=_deduce_type(value))
 
 
 # With some simplifying modifications, this is a shameless copy from
--- a/src/spox/_attributes.py
+++ b/src/spox/_attributes.py
@@ -21,7 +21,7 @@ AttrT = TypeVar("AttrT", bound="Attr")
 AttrIterableT = TypeVar("AttrIterableT", bound="_AttrIterable")
 
 
-class AttributeTypeError(TypeError):
+class OnnxAttributeTypeError(TypeError):
     """Raised upon encountering type errors in node attributes."""
 
     pass
@@ -79,7 +79,7 @@ class Attr(ABC, Generic[T]):
             value_type = f"tuple[{tuple_types}, ...]"
         else:
             value_type = type(self.value).__name__
-        return AttributeTypeError(
+        return OnnxAttributeTypeError(
             f"Unable to instantiate `{type(self).__name__}` with value of type `{value_type}`."
         )
 
--- a/src/spox/_attributes.py
+++ b/src/spox/_attributes.py
@@ -69,13 +69,12 @@ class Attr(ABC, Generic[T]):
 
     def _get_pretty_type_exception(self):
         if isinstance(self.value, tuple) and len(self.value):
-            tuple_types = " | ".join(sorted({type(v).__name__ for v in self.value}))
-            value_type = f"tuple[{tuple_types}, ...]"
+            types = ", ".join(sorted({type(v).__name__ for v in self.value}))
+            msg = f"Unable to instantiate `{type(self).__name__}` from items of type(s) `{types}`."
         else:
-            value_type = type(self.value).__name__
-        return TypeError(
-            f"Unable to instantiate `{type(self).__name__}` with value of type `{value_type}`."
-        )
+            ty = type(self.value).__name__
+            msg = f"Unable to instantiate `{type(self).__name__}` with value of type `{ty}`."
+        return TypeError(msg)
 
 
 class _Ref(Generic[T]):
--- a/src/spox/_attributes.py
+++ b/src/spox/_attributes.py
@@ -76,7 +76,7 @@ class Attr(ABC, Generic[T]):
 
     def _get_pretty_type_exception(self):
         if isinstance(self.value, tuple) and len(self.value):
-            tuple_types = ", ".join({type(v).__name__ for v in self.value})
+            tuple_types = " | ".join(sorted({type(v).__name__ for v in self.value}))
             value_type = f"tuple[{tuple_types}, ...]"
         else:
             value_type = type(self.value).__name__
--- a/src/spox/_attributes.py
+++ b/src/spox/_attributes.py
@@ -32,6 +32,12 @@ AttrT = TypeVar("AttrT", bound="Attr")
 AttrIterableT = TypeVar("AttrIterableT", bound="_AttrIterable")
 
 
+class AttributeTypeError(TypeError):
+    """Raised upon encountering type errors in node attributes."""
+
+    pass
+
+
 class Attr(ABC, Generic[T]):
     _value: Union[T, "_Ref[T]"]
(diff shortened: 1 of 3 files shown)


src/spox/_attributes.py
```diff
@@ -30,6 +32,12 @@
 AttrIterableT = TypeVar("AttrIterableT", bound="_AttrIterable")
 
 
+class AttributeTypeError(TypeError):
```
jbachurski, line 35: Perhaps nitpick: Is it useful to disambiguate the specific TypeError? There isn’t a stable exception type established anyway.

src/spox/_attributes.py
```diff
@@ -188,37 +202,98 @@ def maybe(
 class AttrFloat32s(_AttrIterable[float]):
     _attribute_proto_type_int = AttributeProto.FLOATS
 
-    def _to_onnx_deref(self, key: str) -> AttributeProto:
+    def _to_onnx_deref(self, key: str, only_typecheck: bool = False) -> AttributeProto:
         # ensure values are all floats
         values = [float(v) for v in self.value]
-        return make_attribute(key, values, attr_type=AttributeProto.FLOATS)
+        return (
```
jbachurski, line 208: The difference in behaviour for list attributes should probably be handled in the base class, AttrIterable.

src/spox/_attributes.py
```diff
@@ -188,37 +202,98 @@ def maybe(
...
     return ref._concrete._value
+
+
+def _make_attribute_type(value: Any) -> AttributeProto:
+    return AttributeProto(type=_deduce_type(value))
+
+
+# With some simplifying modifications, this is a shameless copy from
+# https://github.com/onnx/onnx/blob/b60f69412abb5393ab819b936b473f83867f6c87/onnx/helper.py#L838
+# TODO in this PR: we should probably abstract this in onnx and reuse it in onnx's make_attribute; open an issue or PR
```
jbachurski, line 261: This is indeed an uncomfortable situation and I’m not sure what’s a good way about it. Validation constructing a throwaway protobuf til now was a lazy solution that did the job, but is as turns out too inefficient. Maybe there’s other things to piggy-back off of (numpy?)? Perhaps at least the code could be rewritten in a better style rather than pasted.

jbachurski (changes requested):
I don’t think I like the structural changes made here, but I believe I see the goals and they seem right (avoid constructing protobuf and instead have dedicated routines for the type-check). While the most lazy solution would be to do checks lazily at build time, this would deteriorate the Spox experience, so it’s better to find a well-structured and concise way of doing the checks effectively. 
Most of the code here was a result of a refactor by @cbourjau , perhaps he has additional thoughts.

src/spox/_attributes.py
```diff
@@ -82,8 +93,11 @@ def __init__(self, concrete: Attr[T], outer_name: str):
     def copy(self) -> "_Ref[T]":
         return self
 
-    def _to_onnx(self, key: str) -> AttributeProto:
-        parent_type = self._concrete._to_onnx(key).type
+    def _to_onnx(self, key: str, only_typecheck: bool = False) -> AttributeProto:
+        parent_type = self._concrete._to_onnx(
+            key,
+            only_typecheck=True,  # TODO in this PR: make sure we never use the actual attribute proto.
```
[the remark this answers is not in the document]
SimeonStoykovQC: Does `only_typecheck=True` in this context look right to you, i.e. do we ever need to construct a protobuf with values in the `_Ref` case?

src/spox/_attributes.py
SimeonStoykovQC: > Is it useful to disambiguate the specific TypeError?

It's useful if we want to cover those specific error cases, e.g. the test:

https://github.com/Quantco/spox/blob/2c3d882156d4186d2f0d1cda99e27ea0feaf2150/tests/test_standard_op.py#L69-L75

In general, it is encouraged to have custom exception types in Python, because it makes exceptions-for-control-flow less error-prone.

src/spox/_attributes.py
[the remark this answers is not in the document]
cbourjau: `Ref` attributes are only used in the context of ONNX-functions which are very niche in idiomatic Spox usages. I'd be fine simply foregoing any optimization for them at this point in time.

cbourjau (commented):
Let's take a step back. Over the life-cycle of building a model we are likely to not only validate the attributes but to also eventually serialize the `AttributeProto`s into the model. Our problem is that we create them more often then necessary. While optimizing the validation would be beneficial, we also get it for free every time we build the `AttributeProto`.

If we find a clean way to only build the `AttributeProto` once and as early as possible then that would be the best solution. Maybe caching is actually rather straight forward in this case. Correct me if I'm wrong, but could we not simply cache the protobuf object in the `Attr` class (e.g. `Attr._proto: None | AttributeProto`) once we build it the first time for the validation? If we special case [this](https://github.com/Quantco/spox/blob/main/src/spox/_standard.py#L70-L75) step a little bit to only reinstantiate attributes for subgraphs, type protos, and `_Ref`s then we would be fine, wouldn't we?

src/spox/_attributes.py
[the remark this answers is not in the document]
jbachurski: To be completely specific, a Ref attribute can only occur in a FunctionProto which defines a function. A function (when used as an operator) can itself have attributes and operators in its body can refer to them. That's what function attributes do. However, as far as I'm aware, since we never decided to fully support functions, I don't think this support is tested anywhere and this might be dead code.

jbachurski:
> Let's take a step back. Over the life-cycle of building a model we are likely to not only validate the attributes but to also eventually serialize the `AttributeProto`s into the model. Our problem is that we create them more often then necessary. While optimizing the validation would be beneficial, we also get it for free every time we build the `AttributeProto`.
> 
> If we find a clean way to only build the `AttributeProto` once and as early as possible then that would be the best solution. Maybe caching is actually rather straight forward in this case. Correct me if I'm wrong, but could we not simply cache the protobuf object in the `Attr` class (e.g. `Attr._proto: None | AttributeProto`) once we build it the first time for the validation? If we special case [this](https://github.com/Quantco/spox/blob/main/src/spox/_standard.py#L70-L75) step a little bit to only reinstantiate attributes for subgraphs, type protos, and `_Ref`s then we would be fine, wouldn't we?

I think I suggested this originally, but you were concerned about memory usage implications. Have you got thoughts on that? In particular, after the model is built, that memory usage doubles (every attribute will end up copied into the ModelProto – I think).

src/spox/_attributes.py
SimeonStoykovQC: How would you handle the custom behavior in some child classes such as:

https://github.com/Quantco/spox/blob/2c3d882156d4186d2f0d1cda99e27ea0feaf2150/src/spox/_attributes.py#L207
https://github.com/Quantco/spox/blob/2c3d882156d4186d2f0d1cda99e27ea0feaf2150/src/spox/_attributes.py#L241

We rely on exactly the same values & semantics as in the call to `make_attribute` 🤔

src/spox/_attributes.py
SimeonStoykovQC: I'm also unsure what the best approach is. Ideally, we would modularize this upstream, and then reuse the type checking component. Is this realistic?

>  Maybe there’s other things to piggy-back off of (numpy?)

Could you elaborate? We want exactly the same logic as in `make_attribute`, and since the latter is not piggybacking on anything else, I don't see how we can.

> Perhaps at least the code could be rewritten in a better style rather than pasted.

The closer we are to the original implementation, the better. But if you have any concrete suggestions, happy to reflect them.

jbachurski: > I'm also unsure what the best approach is. Ideally, we would modularize this upstream, and then reuse the type checking component. Is this realistic?

Yes, I don’t see issues if one makes the right argument for it upstream. We will have to wait for the next ONNX release, though.

> Could you elaborate? We want exactly the same logic as in make_attribute, and since the latter is not piggybacking on anything else, I don't see how we can.

Yeah, just a random idea - I guess it doesn’t really work out.

> The closer we are to the original implementation, the better. But if you have any concrete suggestions, happy to reflect them.

I would replace the if-statements with a loop (similar as for iterables), and make the raised exceptions more consistent (why are some ValueErrors?). Things like `len(value) == 0` for a list are stylistic, but hurt my soul.

src/spox/_attributes.py
jbachurski: Overridden class attribute/class method?

SimeonStoykovQC: Is https://github.com/Quantco/spox/pull/122/commits/5bad27f2733397916f7a26cefcec5042736b4ae7 what you had in mind? We could also put `_transformed` in the parent, but I was worried it might be a bit confusing at first sight.

src/spox/_attributes.py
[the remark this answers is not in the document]
SimeonStoykovQC: > Perhaps it's okay to remove this partial implementation if it makes the refactoring and optimisations easier?

It doesn't affect it, so I would leave it. Was just curious, thanks!

src/spox/_attributes.py
SimeonStoykovQC: We removed this part (as we won't try this optimisation yet), so I am resolving this conversation.

SimeonStoykovQC:
@cbourjau and I decided to try caching first before implementing this optimization, so this PR ended up being mostly refactoring to support https://github.com/Quantco/spox/pull/119#issuecomment-1827399762

Note this wants to merge in #119 , which is still relevant, as we will always call `make_attribute` in the end.

cbourjau:
> I think I suggested this originally, but you were concerned about memory usage implications. Have you got thoughts on that? 

Yes, I remember our discussions :) At the time the calculus was that creating attributes is so quick that it never matters, but in reality that assumption did not hold. We are not in a situation where the attribute creation is a real pain point, while worries about memory consumption are still abstract. Hence the change of heart.

jbachurski (commented):
Some stylistic comments, but looks fine for the most part.

src/spox/_attributes.py
```diff
@@ -184,38 +200,56 @@ def maybe(
...
 
 class AttrFloat32s(_AttrIterable[float]):
     _attribute_proto_type_int = AttributeProto.FLOATS
 
     def _to_onnx_deref(self, key: str) -> AttributeProto:
-        # ensure values are all floats
-        values = [float(v) for v in self.value]
-        return make_attribute(key, values, attr_type=AttributeProto.FLOATS)
+        try:
+            transformed = tuple(float(v) for v in self.value)
```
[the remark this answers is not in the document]
cbourjau: Also as a minor nitpick: Why are you using a tuple here? Given that we expect to end up with a container with homogeneous element types of dynamic length a `list` seems more appropriate to me.

src/spox/_attributes.py
```diff
@@ -184,38 +200,56 @@ def maybe(
...
-    def _to_onnx_deref(self, key: str) -> AttributeProto:
-        return make_attribute(key, self.value, attr_type=AttributeProto.INTS)
-
 
 class AttrStrings(_AttrIterable[str]):
     _attribute_proto_type_int = AttributeProto.STRINGS
 
     def _to_onnx_deref(self, key: str) -> AttributeProto:
+        try:
+            transformed = tuple(v.encode() for v in self.value)
```
cbourjau, line 229: Same as above. If we provide a `str` then `make_attribute` will call `encode` for us. If something goes wrong then we get an error from onnx/protobuf with a pretty useless error message. At that point, we can either phrase our error message as a suggestion, or we only iterate through all elements if there was indeed an error (which should be very rare) to provide the user with a more definitive error message.

src/spox/_attributes.py
```diff
@@ -184,38 +200,56 @@ def maybe(
...
     _attribute_proto_type_int = AttributeProto.TENSORS
 
     def _to_onnx_deref(self, key: str) -> AttributeProto:
-        tensors = [from_array(t) for t in self.value]
-        return make_attribute(key, tensors, attr_type=AttributeProto.TENSORS)
+        try:
+            transformed = tuple(from_array(v) for v in self.value)
+        except AttributeError as e:
+            raise AttributeTypeError(
+                "Attribute values don't seem to be numpy arrays."
```
cbourjau, line 247: I'm afraid there are plenty of NumPy arrays which will raise an error with `onnx.numpy_helpers.from_array`. Take the following for example:
```python
from_array(np.array(["foo"]))
>>> RuntimeError: Numpy data type not understood yet: dtype('<U3')
```
Come to think of it, we should probably address that issue, too, but my point is that the error message may point the user into the wrong direction.

src/spox/_attributes.py
```diff
@@ -50,20 +46,44 @@ def value(self) -> T:
...
 
+    def _get_pretty_type_exception(self):
+        if isinstance(self.value, tuple) and len(self.value):
+            tuple_types = ", ".join({type(v).__name__ for v in self.value})
+            value_type = f"tuple[{tuple_types}, ...]"
+        else:
+            value_type = type(self.value).__name__
+        return AttributeTypeError(
+            f"Unable to instantiate `{type(self).__name__}` with value of type `{value_type}`."
+        )
```
cbourjau, line 85: I think we can keep it simple at this point. If we tell the user that it was not possible to instantiate ... from the provided values then they probably know what's up. Furthermore, in many cases the onnx-provided error is quite good and we raise it as well. E.g.:
```python
make_attribute("foo", ["bar", ], attr_type=AttributeProto.FLOATS)
```
> TypeError: 'bar' has type str, but expected one of: int, float

src/spox/_attributes.py
```diff
@@ -50,20 +46,44 @@ def value(self) -> T:
...
+
+    @abc.abstractmethod
     def _to_onnx_deref(self, key: str) -> AttributeProto:
         """Conversion method for the dereferenced case."""
         raise NotImplementedError
 
+    def _get_pretty_type_exception(self):
+        if isinstance(self.value, tuple) and len(self.value):
+            tuple_types = ", ".join({type(v).__name__ for v in self.value})
+            value_type = f"tuple[{tuple_types}, ...]"
```
cbourjau, line 80: If we want to be precise, then we should be correct about it! The type of tuple fields is not interchangeable nor can the `...` syntax be used for heterogenous element types (https://docs.python.org/3/library/typing.html#annotating-tuples ). If we want to go down that road we should end up with a variadic tuple with union-type-elements.

src/spox/_attributes.py
cbourjau: IMHO, we should stick with standard error types (i.e. `TypeError`) for now. Introducing more fine-grained errors should better be done project-wide rather than piece-meal and on the fly.

cbourjau (changes requested):
I think we can keep things a bit simpler. Ultimately, the goal should be to ensure that we get a consistent error type across different attribute types with friendly and correct error messages. 

 It seems to me that we can just throw the user input at `make_attribute` and `from_array` and then augment the exception with a friendlier and correct but not necessarily super precise message where necessary. For example, we can say:
"TypeError: Failed to create AttributeProto with type 'FLOATS' from user-provided input."

Furthermore, we raise `from` the original error which is likely to contain more context. In general, I'd think that these two error messages will be sufficient to point the user in the correct direction.

src/spox/_attributes.py
```diff
@@ -50,20 +46,44 @@ def value(self) -> T:
...
 
     def _to_onnx(self, key: str) -> AttributeProto:
         if isinstance(self._value, _Ref):
             return self._value._to_onnx(key)
         return self._to_onnx_deref(key)
 
+    @property
+    @abc.abstractmethod
+    def _attribute_proto_type_int(self) -> int:
+        raise NotImplementedError
```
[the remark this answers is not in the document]
cbourjau: I think `NotImplementedError` (but instantiated) is [fitting](https://docs.python.org/3/library/exceptions.html?highlight=notimplementederror#NotImplementedError) here if we want to go with a property. Otherwise, can't we simply have `_attribute_proto_type_int: int` as a field?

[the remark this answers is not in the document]
SimeonStoykovQC: We get nicer type checks like this:

> 
src/spox/opset/ai/onnx/v17.py:4650: error: Cannot instantiate abstract class "AttrFloat32" with abstract attribute "_attribute_proto_type_int"  [abstract]


image

src/spox/_attributes.py
```diff
@@ -184,38 +200,56 @@ def maybe(
...
     _attribute_proto_type_int = AttributeProto.FLOATS
 
     def _to_onnx_deref(self, key: str) -> AttributeProto:
-        # ensure values are all floats
-        values = [float(v) for v in self.value]
-        return make_attribute(key, values, attr_type=AttributeProto.FLOATS)
+        try:
+            transformed = tuple(float(v) for v in self.value)
+        except ValueError as e:
+            raise AttributeTypeError("Attribute values don't seem to be floats.") from e
```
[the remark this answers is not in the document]
SimeonStoykovQC: > Also, the non-imperative language makes it sound like the real cause of the exception is unknown

It's more of a best effort for now since I'm not checking the exception's message.

> Why only catch ValueError, as in float("abc"), but not TypeError, as in float([])

👍 

> Why not use _get_pretty_type_exception here? Same applies to AttrStrings.

`_get_pretty_type_exception` is more specialized to the validation. In practice, we should indeed reach `_to_onnx_deref` from the `_validate` call first, but it's also called later on, so "Unable to instantiate `{type(self).__name__}` with value of type `{value_type}`." won't apply that well anymore.

src/spox/_attributes.py
SimeonStoykovQC: This will make me happier too :) Check https://github.com/Quantco/spox/pull/122/commits/09df28f1faa2539e1bf60cd2ddf595b092192925

src/spox/_attributes.py
SimeonStoykovQC: It does provide some more context, while also keeping the given one, e.g. the concrete attr class and all the types in the tuple. It also makes the errors more consistent.

src/spox/_attributes.py
SimeonStoykovQC: It's really not a big deal and it's usually done a lot and considered a good practice. Here it extends `TypeError`, and its name suggests it is coupled with onnx attributes. Let's stick with it, it makes exception handling easier.

cbourjau: From the user's point of view, all public functions from the opset are regular Python functions with only `spox.Var`, built-in, NumPy types in their signature. The concept of an "Attribute" being constructed behind the scenes is an implementation detail that does not need to be part of the public interface. Conceptually, writing `op.constant(value_ints=["a"])` is exactly the same error as writing `range("foo")`.

src/spox/_attributes.py
```diff
@@ -50,20 +46,44 @@ def value(self) -> T:
...
-            )
+        try:
+            type_in_onnx = self._to_onnx_deref("dummy").type
+        except TypeError as e:
+            # TypeError: 'a' has type str, but expected one of: int -- this is raised from within protobuf
+            # when .extend()-ing with the wrong type.
+            if "but expected one of" in str(e):
+                raise self._get_pretty_type_exception() from e
+            else:
+                raise
```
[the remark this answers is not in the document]
cbourjau: TIL: https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement 👍

cbourjau (approved):

PR #123, opened 2023-12-01
Caching for attributes that allow it

SimeonStoykovQC:
# Checklist


src/spox/_standard.py
```diff
@@ -65,13 +66,26 @@ def to_singleton_onnx_model(
         # We inject the evaluated attribute values here and then substitute back
         self_attrs = self.attrs
         try:
+            current_fields = self_attrs.get_fields().items()
```
cbourjau, line 69: It would be great to keep the caching logic contained in the `_attributes.py` file. I think the previous code had the intention of "dereferencing" the `value` here. Maybe we could add a `deref` method to the `Attr` class which creates a new `Attr` instance if `self` is a `_Ref` and `self` otherwise.

cbourjau: I.e. rather than calling `type(v)(v.value)` as done now, we would call `v.deref()`

cbourjau:
> @cbourjau does this look legit? How can I ensure correctness (up for a pair-testing session!)? I'm seeing ~2x speedup in an end-to-end test that constructs, builds, and serializes a graph.

That sounds very promising and looks like a great start! Assuming that the user never mutates these private objects, which correctness issues do you anticipate? I am certainly happy to have a pair-programming session!

jbachurski:
If you are worried about correctness, it would be sufficient to check if the cached AttributeProto has the same attribute name as requested, and otherwise it can be copied and a new name can be set on the copy. You shouldn't see any copies, and if they are necessary in some corner cases, they will be observable as a slight performance regression (perhaps a warning can be added if wanted). Maybe something like this would work:
```py
         if self._last_onnx is None:
             self._last_onnx = self._to_onnx_deref(key)
         if self._last_onnx != key:
             proto = onnx.AttributeProto()
             proto.CopyFrom(self._last_onnx)
             proto.name = key
             self._last_onnx = proto
        return self._last_onnx
```

Why are the changes in `_standard` necessary? I'm not sure I can tell.
If `_Ref` is too annoying to handle (and I could see why that would be...) perhaps it should be removed, as we do not use functions anywhere (they are not that useful in the first place).

cbourjau:
> Why are the changes in _standard necessary?

Right now we are creating new `Attr` objects with `type(v)(v.value)` which would strip the cache away.

SimeonStoykovQC:
> it would be sufficient to check if the cached AttributeProto has the same attribute name as requested, and otherwise it can be copied and a new name can be set on the copy.

The key is different when type checking:

https://github.com/Quantco/spox/blob/d59a9998c7158f00ec17334ad4d4210402e0095b/src/spox/_attributes.py#L53

As you can see, I'm completely ignoring it for now:

https://github.com/Quantco/spox/blob/0f109a17b10ccdf0c82f9c7d95d297986be46ebe/src/spox/_attributes.py#L61-L67

Do you see any concrete problems with this, and can we add a test?

cbourjau:
Good point about the "dummy" keys used during the validation. The only worry I have about mutating the attribute is if its reference is hanging around elsewhere after a build call. Attributes will become part of Nodes when building. Thus, a test that reflects the real world would be if the attribute's name is updated in a `NodeProto` object:

```python
import spox
from onnx import AttributeProto
from onnx.helper import make_node

attr = spox._attributes.AttrFloat32s([1., 2.])
attr_onnx = attr._to_onnx("foo")

node = make_node("FooNode", inputs=["bar"], outputs=["baz"])
node.attribute.append(attr_onnx)

assert node.attribute[0].name == "foo"

attr_onnx.name = "ZZZ"
assert node.attribute[0].name == "foo"
```

The result is that `NodeProto.append` appears to make a copy so we are in the clear in this regard, but should add it as a test to catch any future breakage upstream. We may also add a stern warning to the docstring of `_to_onnx` that the return value of that function should not be assigned to anything long-lived and is primarily intended to be used immediately in `NodeProto.append` (which makes a copy as shown above).

cbourjau:
For future reference: Simeon and I just discussed this issue. It is a bit scary to mutate the cached value with a user-provided name and it would be great to avoid doing that. To that end, we realized that the name is usually known from the start when creating the `Attrxxx` object in the opset. We can alter the opset generation to pass the key there and that should safe us from mutating it later. Only thing that may require a little more thought is where which name belongs for `_Ref` attributes.

src/spox/_standard.py
SimeonStoykovQC: Is https://github.com/Quantco/spox/pull/123/commits/12573b5f2849ef1a8f605f0c6d0a7f76cba791fc what you had in mind?

src/spox/_attributes.py
```diff
@@ -32,15 +32,28 @@
...
         self._value = value
+        self._name = name
+        self._cached_onnx = None
+
         self._validate()
 
+    def recreate_if_not_cachable(self, name: str) -> "Attr":
+        if isinstance(self, (AttrGraph, AttrType, _Ref)) or isinstance(
+            self._value, _Ref
+        ):
```
SimeonStoykovQC, line 49: This looks a bit suspicious to me now. Can we not have infinite-depth nesting? As in, should we also not check `self._value._value` and so on?

cbourjau: It is rather convoluted... The property `self.value` (used below) will call `_deref` which will recurse. I think this is ok?

src/spox/_standard.py
cbourjau: IIUC, calling `type(v)(v.value)` is essentially meant to deref. If so, we may add an `Attr.deref` method that returns `self` if `v._value` is not `_Ref` and otherwise returns `type(self)(self.value)`.

cbourjau: The original code here was only concerned with dereferencing `_Ref`s. The way it was written, it stripped away our cache, but that was just an accident. 

It is unrelated to this line in the code, but note that we still cannot cache the Protobuf object for `AttrGraph` objects because we acutally never build it in `_attributes.py` in the first place: https://github.com/onnx/onnx/blob/main/docs/Operators.md#reshape

SimeonStoykovQC:
With caching:
Screenshot 2023-12-14 at 13 14 09

Without caching:
Screenshot 2023-12-14 at 13 14 37

```python
import numpy as np
import onnx
import spox
import spox.opset.ai.onnx.ml.v3 as ml

Strings = spox.Tensor(np.str_, ("K",))
s = spox.argument(Strings)

NUM_ELEMENTS = 15_000_000

encoded = ml.label_encoder(
    s,
    keys_strings=[str(idx) for idx in range(NUM_ELEMENTS)],
    values_floats=[float(idx) for idx in range(NUM_ELEMENTS)],
)

onnx_model = spox.build(
    inputs={"s": s},
    outputs={"encoded": encoded},
)

onnx.save(onnx_model, f"model_spox.onnx", save_as_external_data=True)
```

cbourjau (approved):
Nice! Thanks for seeing this through! I think a performance boost like this warrants a mention in the changelog :) Can you add that please and then we are ready to merge!

CHANGELOG.rst
```diff
@@ -17,6 +17,7 @@ Change log
 **Other changes**
 
 - The validation of Node attributes has been improved and more consistent exceptions are raised if needed.
+- ONNX node attributes are now computed only once and then cached so that the values are reused for validation and building the model.
```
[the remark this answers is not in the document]
SimeonStoykovQC: Providing more details could be a good lead in debugging, especially if it's about "one of the hardest problems" - caching :)

SimeonStoykovQC:
## Memory benchmarks

I tried to memory benchmark with `memory_profiler`, but the results could vary a lot. Still, things look comparable:

### With caching:
```
(spox) ➜  playin git:(main) ✗ python -m memory_profiler simi_spox.py
Filename: simi_spox.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    15    636.0 MiB    636.0 MiB           1   @profile
    16                                         def main():
    17   1928.5 MiB   1292.5 MiB           1       encoded = ml.label_encoder(s, keys_strings=strings, values_floats=floats)
    18   2246.6 MiB    318.1 MiB           1       encoded2 = ml.label_encoder(encoded, keys_floats=floats, values_strings=strings)
    19   2653.4 MiB    406.7 MiB           1       encoded3 = ml.label_encoder(encoded2, keys_strings=strings, values_floats=floats)
    20   3590.9 MiB    937.6 MiB           1       onnx_model = spox.build(inputs={"s": s}, outputs={"encoded3": encoded3})
    21   3400.5 MiB   -190.4 MiB           1       onnx.save(onnx_model, f"model_spox.onnx", save_as_external_data=True)
    22   3400.5 MiB      0.0 MiB           1       pass


(spox) ➜  playin git:(main) ✗ python -m memory_profiler simi_spox.py
Filename: simi_spox.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    15    635.4 MiB    635.4 MiB           1   @profile
    16                                         def main():
    17   1928.6 MiB   1293.2 MiB           1       encoded = ml.label_encoder(s, keys_strings=strings, values_floats=floats)
    18   2254.8 MiB    326.2 MiB           1       encoded2 = ml.label_encoder(encoded, keys_floats=floats, values_strings=strings)
    19   2661.5 MiB    406.7 MiB           1       encoded3 = ml.label_encoder(encoded2, keys_strings=strings, values_floats=floats)
    20   3135.7 MiB    474.2 MiB           1       onnx_model = spox.build(inputs={"s": s}, outputs={"encoded3": encoded3})
    21   2746.3 MiB   -389.4 MiB           1       onnx.save(onnx_model, f"model_spox.onnx", save_as_external_data=True)
    22   2746.3 MiB      0.0 MiB           1       pass
```

### Without caching

```
(spox) ➜  playin git:(main) ✗ python -m memory_profiler simi_spox.py
Filename: simi_spox.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    15    636.0 MiB    636.0 MiB           1   @profile
    16                                         def main():
    17   1673.0 MiB   1037.0 MiB           1       encoded = ml.label_encoder(s, keys_strings=strings, values_floats=floats)
    18   1810.7 MiB    137.7 MiB           1       encoded2 = ml.label_encoder(encoded, keys_floats=floats, values_strings=strings)
    19   2029.0 MiB    218.3 MiB           1       encoded3 = ml.label_encoder(encoded2, keys_strings=strings, values_floats=floats)
    20   2882.0 MiB    853.0 MiB           1       onnx_model = spox.build(inputs={"s": s}, outputs={"encoded3": encoded3})
    21   2787.9 MiB    -94.1 MiB           1       onnx.save(onnx_model, f"model_spox.onnx", save_as_external_data=True)
    22   2787.9 MiB      0.0 MiB           1       pass


(spox) ➜  playin git:(main) ✗ python -m memory_profiler simi_spox.py
Filename: simi_spox.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    15    635.9 MiB    635.9 MiB           1   @profile
    16                                         def main():
    17   1685.9 MiB   1050.0 MiB           1       encoded = ml.label_encoder(s, keys_strings=strings, values_floats=floats)
    18   1815.6 MiB    129.7 MiB           1       encoded2 = ml.label_encoder(encoded, keys_floats=floats, values_strings=strings)
    19   2033.9 MiB    218.3 MiB           1       encoded3 = ml.label_encoder(encoded2, keys_strings=strings, values_floats=floats)
    20   2960.3 MiB    926.4 MiB           1       onnx_model = spox.build(inputs={"s": s}, outputs={"encoded3": encoded3})
    21   2781.6 MiB   -178.8 MiB           1       onnx.save(onnx_model, f"model_spox.onnx", save_as_external_data=True)
    22   2781.6 MiB      0.0 MiB           1       pass


(spox) ➜  playin git:(main) ✗ python -m memory_profiler simi_spox.py
Filename: simi_spox.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    15    628.8 MiB    628.8 MiB           1   @profile
    16                                         def main():
    17   1672.7 MiB   1044.0 MiB           1       encoded = ml.label_encoder(s, keys_strings=strings, values_floats=floats)
    18   1818.5 MiB    145.8 MiB           1       encoded2 = ml.label_encoder(encoded, keys_floats=floats, values_strings=strings)
    19   2036.9 MiB    218.3 MiB           1       encoded3 = ml.label_encoder(encoded2, keys_strings=strings, values_floats=floats)
    20   3013.8 MiB    977.0 MiB           1       onnx_model = spox.build(inputs={"s": s}, outputs={"encoded3": encoded3})
    21   2833.6 MiB   -180.2 MiB           1       onnx.save(onnx_model, f"model_spox.onnx", save_as_external_data=True)
    22   2833.6 MiB      0.0 MiB           1       pass

```

CHANGELOG.rst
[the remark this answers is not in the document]
cbourjau: Fair enough!

cbourjau:
Each block in the benchmark shows results from re-running the same test? Either way, doesn't look as if things are out of control (which I didn't expect anyhow). I think this is good to go! Thanks!

PR #126, opened 2023-12-07
Bind name to attributes upon construction

SimeonStoykovQC:
See https://github.com/Quantco/spox/pull/123#issuecomment-1845282902 for more context. Belongs to #123 but wanted to hear everyone's thoughts and separate the changes to make them easily reviewable.

# Checklist


src/spox/_graph.py
```diff
@@ -37,17 +37,23 @@ def arguments_dict(**kwargs: Optional[Union[Type, numpy.ndarray]]) -> Dict[str,
     """
     result = {}
     for name, info in kwargs.items():
-        attr_name = AttrString(name)
+        attr_name = AttrString(value=name, name="dummy")
         if isinstance(info, Type):
             result[name] = Argument(
-                Argument.Attributes(name=attr_name, type=AttrType(info), default=None),
+                Argument.Attributes(
```
cbourjau, line 43: As discussed, it is a bit unfortunate that these "internal nodes" reuse that clunky interface that, but that is certainly out of scope to change in this PR.

src/spox/_node.py
```diff
@@ -373,7 +373,8 @@ def to_onnx(
                     subgraph = build_subgraph(self, key, attr.value)
                     attr_proto = onnx.helper.make_attribute(key, subgraph)
                 else:
-                    attr_proto = attr._to_onnx(key)
+                    assert attr._name == key  # TODO-in-this-PR: remove
```
cbourjau, line 376: Adding an explicit comment so that we won't forget

cbourjau (commented):
I still think this is the way to go. Any reason you keep this as a draft?

adityagoel4512:
This looks good to me as well.

CHANGELOG.rst
```diff
@@ -14,6 +14,10 @@ Change log
 
 - The opsets ``ai.onnx`` version 20 and ``ai.onnx.ml`` version 4 (ONNX 1.15) are now shipped with Spox.
 
+**Other changes**
+
+- Internal spox classes of node attributes now receive the attribute's name earlier, in the constructor, instead of passing it later when building the attribute proto.
+
```
jbachurski, line 20: Unsure what others think - but I don't think this is important in a changelog. It does have a place in the commit history and that's probably enough, as it should just be an internal refactor.

jbachurski (approved):
Approach looks fine. Things are a bit more redundant (e.g. a node holds an attribute of a given name which itself also has a name), but this lack of redundancy did inhibit optimisations so it's fair to do.

CHANGELOG.rst
adityagoel4512: I agree, the changelog should document the impact for the user. This can be rolled into improved caching/performance once merged to the main branch.

PR #134, opened 2024-02-07
Imporve ONNX 1.14 compatibility

cbourjau:
ONNX 1.15 introduced some performance optimizations which we are using since `0.10`. We can forego their usage if `onnx<1.15` without loss of functionality to avoid a version bound. 

Since Spox exports the ONNX standard which expands with each (minor) release, we can't claim full compatibility with older versions.

# Checklist

### DIFF
--- a/src/spox/_attributes.py
+++ b/src/spox/_attributes.py
@@ -210,9 +210,16 @@ class _AttrIterable(Attr[Tuple[S, ...]], ABC):
         return cls(tuple(value), name) if value is not None else None
 
     def _to_onnx_deref(self) -> AttributeProto:
-        return make_attribute(
-            self._name, self.value, attr_type=self._attribute_proto_type
-        )
+        # the attr_type argument was only added in onnx>=1.15. It is a
+        # performance optimization for which we don't want to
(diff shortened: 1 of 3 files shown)


src/spox/_attributes.py
```diff
@@ -210,9 +210,16 @@ def maybe(
...
 
     def _to_onnx_deref(self) -> AttributeProto:
-        return make_attribute(
-            self._name, self.value, attr_type=self._attribute_proto_type
-        )
+        # the attr_type argument was only added in onnx>=1.15. It is a
+        # performance optimization for which we don't want to
+        # introduce a version bound. If we fail, we try again without
+        # it.
+        try:
```
adityagoel4512, line 217: Could we explicitly check the ONNX version instead of relying on catching TypeError? TypeErrors could be raised for [different](https://github.com/onnx/onnx/blob/48501e0ff59843a33f89e028cbcbf080edd0c1df/onnx/helper.py#L972) reasons which should surface here.

adityagoel4512 (changes requested):

adityagoel4512 (approved):

PR #135, opened 2024-02-07
Add packaging dependency

cbourjau:
This should have been added with #134 . The pypi package is currently broken if used in isolation. It pains me to pull in a new dependency, but it is a very light one (no secondary dependencies), it is the de facto standard, and I prefer this over hacking my own version string parsing. 

# Checklist

### DIFF
--- a/.github/workflows/build_and_publish.yml
+++ b/.github/workflows/build_and_publish.yml
@@ -1,6 +1,7 @@
 name: Build and upload to PyPI
 
 on:
+  push:
   pull_request:
   release:
     types:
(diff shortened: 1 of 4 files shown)


adityagoel4512:
How was this not surfaced in the CI before?

cbourjau:
I now added a step to the CI that installs and imports the package in a clean environment. We didn't run into the issue because `packaging` is a very common dependency that was installed through a transient dependency of one of our dev-dependencies. `pytest` for instance depends on it: https://prefix.dev/channels/conda-forge/packages/pytest

adityagoel4512 (approved):

PR #138, opened 2024-02-14
Improve performance

cbourjau:
We currently store a traceback for every node we create, but there is no public way to expose that information. This PR foregoes the creation of the traceback based on a (private) global variable which may be set when needed. This change provides a considerable speedup for the default configuration.

# Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ Change log
 
 **Other changes**
 
-- Improved performance
+- Improved node creation speed by skipping the storing of the traceback
 
 
 0.10.2 (2023-02-08)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@
 Change log
 ==========
 
+0.10.3 (unreleased)
+-------------------
+
+**Other changes**
+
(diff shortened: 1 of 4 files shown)


adityagoel4512:
From external chat it appears this is for onnxruntime debugging only? If so, while nice to see where the Spox node was constructed that results in the failure, you should be able to determine which operator is the culprit from a runtime stack trace. This is only used internally so I'd be happy to remove it, but worth checking with @jbachurski who will know more.

CHANGELOG.rst
```diff
@@ -7,6 +7,14 @@
 Change log
 ==========
 
+0.10.3 (unreleased)
+-------------------
+
+**Other changes**
+
+- Improved performance
```
adityagoel4512, line 15:
```suggestion
- Improved node creation speed by skipping the storing of the traceback
```

jbachurski:
Trackbacks are used so that if only a node name is available in an error message, you get a hint what node corresponds to it. This usually would happen at runtime, hence why there is something for ORT implemented. I'm happy for it to be disabled by default since these inspections can be rather slow.

adityagoel4512 (approved):

PR #141, opened 2024-02-29
Use ruff instead of black, isort, and flake8

cbourjau:
This PR replaces black, isort, and flake8 with ruff. Ruff does a great job removing unused imports in our opsets. There are some other minor formatting changes that I think are not worth the trouble of diverging from the default settings. 

Supersedes #133

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -29,7 +29,7 @@ class ONNXRuntimeHelper:
         debug_index = {
             **graph._get_build_result().scope.var.of_name,
             **graph._get_build_result().scope.node.of_name,
-        }  # type: ignore
+        }
         if self._last_graph is graph:
             print("[ONNXRuntimeHelper] Reusing previous session.", file=sys.stderr)
             session = self._last_session
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -26,7 +26,10 @@ class ONNXRuntimeHelper:
         self._last_session = None
 
     def run(self, graph: Graph, unwrap: Optional[str] = None, **kwargs):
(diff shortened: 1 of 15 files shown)


jbachurski:
Looks good 🚀 
However, I'm not super familiar with the ruff configuration details, maybe someone else can take a look to approve.

cbourjau:
@adityagoel4512 Would you like to do the honors unless you see something amiss? Given that the config is quite defaulty and the diff is reasonably small I'd be happy to just see how it works out.

tests/conftest.py
```diff
@@ -26,7 +26,10 @@ def __init__(self):
         self._last_session = None
 
     def run(self, graph: Graph, unwrap: Optional[str] = None, **kwargs):
-        debug_index = {**graph._get_build_result().scope.var.of_name, **graph._get_build_result().scope.node.of_name}  # type: ignore
+        debug_index = {
+            **graph._get_build_result().scope.var.of_name,
+            **graph._get_build_result().scope.node.of_name,
+        }  # type: ignore
```
adityagoel4512, line 32:
```suggestion
        }
```

adityagoel4512 (approved):
Not really in the scope of this PR but it would be great if we could address the attribute reference type hinting (e.g. by updating the opset generation).

cbourjau:
The attribute references are still private which is the reason why they are not in the public interface. On the other hand, maybe it is time to stabilize ONNX functions in Spox now that their support in the onnxruntime seems to have matured (even though I consider them pretty broken and useless in the standard).

PR #143, opened 2024-03-05
Support shallow copy of Var objects and raise error on deepcopy

### LINKED ISSUE #139: spox.Var cannot be copied
Making a shallow copy of a `spox.Var` object leads to an error at build time. While it seems reasonable that a `copy.deepcopy` would not work, it seems surprising from a user perspective that a shallow copy is also causing issues. 

**[Failing test case](https://github.com/Quantco/spox/commit/db65f8dc876757720de151a054e04f271d17b9bb):**

```python
def test_copy_vars():
    from copy import copy

    a = argument(Tensor(float, ()))
    b = op.add(a, copy(a))
    build({"a": a}, {"b": b})
```

**Stack trace:**
```
    def test_copy_vars():
        from copy import copy
        a = argument(Tensor(float, ()))
        b = op.add(a, copy(a))
>       build({"a": a}, {"b": b})

tests/test_public.py:95: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/spox/_public.py:117: in build
    return graph.to_onnx_model()
src/spox/_graph.py:390: in to_onnx_model
    opsets = self.get_opsets()
src/spox/_graph.py:249: in get_opsets
    return max_opset_policy(self._get_opset_req())
src/spox/_graph.py:259: in _get_opset_req
    return self._get_build_result().opset_req | (
src/spox/_graph.py:254: in _get_build_result
    self._build_result.value = _build.Builder(self).build_main()
src/spox/_build.py:203: in build_main
    return self.compile_graph(self.main, Scope())
src/spox/_build.py:456: in compile_graph
    nodes[node] = tuple(node.to_onnx(scope, build_subgraph=build_subgraph))
src/spox/_node.py:354: in to_onnx
    input_names = [scope.var[var] if var is not None else "" for var in self.inputs]
src/spox/_node.py:354: in <listcomp>
    input_names = [scope.var[var] if var is not None else "" for var in self.inputs]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <spox._scope.ScopeSpace object at 0x10febc390>
item = <Var from spox.internal@0::Argument->?? of float64>

    def __getitem__(self, item: Union[str, H]):
        """Access the name of an object or an object with a given name in this (or outer) namespace."""
        if self.parent is not None and item in self.parent:
            return self.parent[item]
        elif isinstance(item, str):
            return self.of_name[item]
        else:
>           return self.name_of[item]
E           KeyError: <Var from spox.internal@0::Argument->?? of float64>

src/spox/_scope.py:76: KeyError
```

In any case, we should consider adding a [`__deepcopy__`](https://docs.python.org/3/library/copy.html#copy.deepcopy) method to `Var` which raises a more helpful exception immediately.

cbourjau:
Closes #139  by returning self for a shallow copy and raising an error for deep copies

# Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ Change log
 **Other changes**
 
 - Improved node creation speed by skipping the storing of the traceback
-- :class:`spox.Var` objects may now be shallow copied
+- :class:`spox.Var` objects may now be shallow copied. Deep copies are explicitly prohibited.
 
 
 0.10.2 (2023-02-08)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Change log
 **Other changes**
 
 - Improved node creation speed by skipping the storing of the traceback
+- :class:`spox.Var` objects may now be shallow copied
 
 
 0.10.2 (2023-02-08)
--- a/src/spox/_var.py
+++ b/src/spox/_var.py
@@ -131,6 +131,14 @@ class Var:
         """Equivalent to ``self.unwrap_type().unwrap_optional()``."""
         return self.unwrap_type().unwrap_optional()
 
+    def __copy__(self) -> "Var":
+        # Simply return `self` to ensure that "copies" are still equal
+        # during the build process
+        return self
+
+    def __deepcopy__(self, _) -> "Var":
+        raise ValueError("'Var' objects cannot be deepcopied.")
+
     def __add__(self, other) -> "Var":
         return Var._operator_dispatcher.add(self, other)
 
--- a/tests/test_public.py
+++ b/tests/test_public.py
@@ -1,3 +1,5 @@
+from copy import copy, deepcopy
+
 import numpy as np
 import onnx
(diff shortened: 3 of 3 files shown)


CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ Change log
 **Other changes**
 
 - Improved node creation speed by skipping the storing of the traceback
+- :class:`spox.Var` objects may now be shallow copied
```
adityagoel4512, line 16:
```suggestion
- :class:`spox.Var` objects may now be shallow copied. Deep copies are explicitly prohibited.
```

adityagoel4512 (approved):

Issue #145, opened 2024-03-11
Node adaptation fails when composing inlined model with node with subgraph.


adityagoel4512:
The described error can be reproduced with the following example:

```python
@pytest.fixture
def old_squeeze() -> onnx.ModelProto:
    return onnx.parser.parse_model(
        """
<
 ir_version: 8,
 opset_import: ["" : 12]
>
agraph (float[1, N] A) => (float[N] B)
{
    B = Squeeze<axes = [0]>(A)
}
"""
    )

def test_if_adaptation(old_squeeze):
        a = argument(Tensor(numpy.float32, (1, None,)))
    sq = inline(old_squeeze)(A=a)["B"]
    b = argument(Tensor(numpy.float32, ("N",)))
    cond = op19.equal(sq, b)
    out = op18.if_(cond,
        then_branch=lambda: [sq],
        else_branch=lambda: [sq])
    build({"a": a, "b": b}, {"out": out[0]})
```

Spox (correctly) attempts to update the opset of the if node from 18 to 19, but in doing so is unable to resolve the input name:

```
tests/test_adapt.py:121:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/spox/_public.py:117: in build
    return graph.to_onnx_model()
src/spox/_graph.py:412: in to_onnx_model
    self.to_onnx(concrete=concrete),
src/spox/_graph.py:343: in to_onnx
    node_protos = itertools.chain.from_iterable(self.get_adapted_nodes().values())
src/spox/_graph.py:281: in get_adapted_nodes
    best_effort = adapt_best_effort(
src/spox/_adapt.py:142: in adapt_best_effort
    adapted = adapt_node(
src/spox/_adapt.py:60: in adapt_node
    onnx.checker.check_model(source_model, full_check=True)
[...]

PR #146, opened 2024-03-11
Bail node adaptation if node has subgraph

adityagoel4512:
We should avoid node adaptation when the node in question has a subgraph (until we have a proper implementation that addresses issues like https://github.com/Quantco/spox/issues/145). This would affect the "If", "Loop", "Scan" and "SequenceMap" nodes. 

Note that previously, this would succeed if the subgraph did not reference anything outside of it. Is it worth making this check more complex/fine-grained and allow these to continue to succeed?

# Checklist


adityagoel4512:
Do you have any concerns with this PR @cbourjau?

cbourjau (approved):

Issue #140, opened 2024-02-28
Naming algorithm has quadratic time complexity


cbourjau:
The [current naming algorithm](https://github.com/Quantco/spox/blame/main/src/spox/_scope.py#L121) is defined as:

```python
  def enum(self, base: str, suffix: str = "_{}") -> str:
      """Find an unused name by enumerating the pattern ``base + suffix.format(i)`` through `i = 0, 1, ...`"""
      i = 0
      while (name := f"{base}{suffix.format(i)}") in self:
          i += 1
      return name
```

This function becomes very expensive in (contrived) graphs where the same `base` occurs very often (1000+ times). We may consider simply having a single index across all nodes and edges.

jbachurski:
Ah yes, the original (ah, so very old by now!) build code did take care of this. IIRC, it used to do something like this:

```py
key = (base, suffix)
i = self.memo_enum.setdefault(key, 0)
while (name := f"{base}{suffix.format(i)}") in self:
    i += 1
self.memo_enum[key] = i
return name
```
To be honest, there's no real use for different suffixes I think, but the above should be robust to them.  This doesn't increase space complexity as we are already keeping track of all the names.

But maybe this isn't the simplest. Here's some other solutions:

* Use the common index, like you say. That's probably fine.
* Get rid of base altogether (depends if you think the improved generated code 'quality' is worth it)
* Get rid of the custom suffixes so finding the next free index isn't as weird. Technically with bad `suffix`es it's possible for this code to loop forever...

cbourjau:
I think your code snippet is perfectly fine! Thanks!

PR #154, opened 2024-05-17
Optimize the naming algorithm in scopes

aivanoved:
Solves #140

# Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@
 Change log
 ==========
 
+0.12.1 (2024-XX-XX)
+-------------------
+
+**Other changes**
+
+- Imporved quadratic complexity of the naming algorithm in the build process, by keeping track of how many times a base name was used in current and all containing scopes.
+
+
 0.12.0 (2024-05-16)
 -------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@
 Change log
 ==========
 
-0.12.1 (2024-XX-XX)
+(unreleased)
 -------------------
 
 **Other changes**
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,12 +7,12 @@
 Change log
 ==========
 
-(unreleased)
+0.12.1 (unreleased)
 -------------------
 
 **Other changes**
 
-- Imporved quadratic complexity of the naming algorithm in the build process, by keeping track of how many times a base name was used in current and all containing scopes.
+- The node-naming algorithm now has constant rather than quadratic time complexity.
 
 
 0.12.0 (2024-05-16)
--- a/src/spox/_scope.py
+++ b/src/spox/_scope.py
@@ -123,7 +123,7 @@ class ScopeSpace(Generic[H]):
         del self.name_of[value]
 
     def enum(self, base: str) -> str:
-        """Find an unused name by enumerating the pattern ``base + suffix.format(i)`` through `i = 0, 1, ...`"""
+        """Find an unused name by enumerating the pattern ``f"{base}_{i}"`` through `i = 0, 1, ...`"""
         self.base_name_counters.setdefault(base, 0)
 
         name = f"{base}_{self.base_name_counters[base]}"
--- a/src/spox/_scope.py
+++ b/src/spox/_scope.py
@@ -23,7 +23,7 @@ class ScopeSpace(Generic[H]):
     name_of: Dict[H, str]
     of_name: Dict[str, H]
     reserved: Set[str]
-    name_resolution: Dict[str, int]
+    base_name_counters: Dict[str, int]
     parent: "Optional[ScopeSpace[H]]"
 
     def __init__(
@@ -53,7 +53,9 @@ class ScopeSpace(Generic[H]):
         # all scopes.
         # While the standard is more lenient in this respect, we
         # simply don't allow any name reuse across the entire model.
-        self.name_resolution = parent.name_resolution if parent is not None else dict()
+        self.base_name_counters = (
+            parent.base_name_counters if parent is not None else dict()
+        )
 
     def __contains__(self, item: Union[str, H]) -> bool:
         """Checks if a given name or object is declared in this (or outer) namespace."""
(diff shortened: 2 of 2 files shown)


aivanoved:
Thank you for the quick review,

I understand where your concerns come from. In fact, while deciding on whether this is the approach I wanted to take, I did share some of them. I'll be addressing them in what I believe is reverse order of severity

### Point 3

> The code changed more than it seems necessary - preferably this just addresses Naming algorithm has quadratic time complexity #140 algorithmically as outlined by the issue, and then a refactor can be done if that's helpful.

This is a consequence of all the other points so, if we implement them in the way suggested by the review I believe it will be automatically solved. Nonetheless, this is valid criticism as needless code changes and technical depth are definitely not a great idea

I do however still think, this change is not unnecessary in the way it is currently written in the pr

### Point 2

> Abusing Python's iterator interface just to write next(suffix) vs suffix.next()

I don't think this is abuse of the interface. What we are creating is a generator of suffixes and is by that definition an iterator. In fact I see the use of `suffix.next()` as even more problematic as it does not imply interior mutation of the `Suffix` object, whereas `next(suffix)` does.

### Point 1

> Introduction of a new class, with an ABC despite there being one implementation and this being an internal utility.

The introduction of a base class is purely for typing and clarity purposes, it is essentially an interface nothing more than a marker. The reason we need this marker class (yes, we can definitely argue this might be too dogmatic of an OOP approach here) is that, otherwise there is no other clear way of specifying what type of objects are acceptable to be passed to the `enum` and `maybe_enum` functions (`Iterator[str]` is an option, but that does not carry enough information in my opinion to describe the idea of that object)

We need a base/marker class if we want to be able to pass different `Suffix` classes to the function not just the default one if we make it just one class then the argument of `suffix` becomes obsolete. This is definitely a simplifying approach, and one which I would be in favour of.

### Point 4

> This moves the dictionary into integers I proposed in Naming algorithm has quadratic time complexity #140 into another dictionary into ref-cells of integers, plus there's a redundancy where the Suffix class stores the name and the dictionary indexes by those names. A bunch of abstractions that don't feel useful right now.

I am a little confused as to what you mean in the second to last sentence here.

In the proposal in #140 for each base name we put into the `memo_enum` dictionary each `(base, suffix)` tuple as a key, here we have the dictionary storing only the one and only `Suffix` iterator object for the base name in all scopes, current or parent, so in fact there is much less redundancy by the approach in this pr

# Note on the proposal in #140

In any case, I understand why this might seem over engineered. I would not be opposed to implementing the algo as proposed in #140

src/spox/_scope.py
```diff
@@ -23,6 +23,7 @@ class ScopeSpace(Generic[H]):
     name_of: Dict[H, str]
     of_name: Dict[str, H]
     reserved: Set[str]
+    name_resolution: Dict[str, int]
```
cbourjau, line 26: How about:
```suggestion
    base_name_counters: Dict[str, int]
```
?

src/spox/_scope.py
```diff
@@ -48,6 +49,7 @@ def __init__(
         self.of_name = of_name.copy() if of_name is not None else {}
         self.reserved = reserved.copy() if reserved is not None else set()
         self.parent = parent
+        self.name_resolution = parent.name_resolution if parent is not None else dict()
```
cbourjau, line 52:
```suggestion
        # Reference to a single `base_name_counters` object across
        # all scopes.
        # While the standard is more lenient in this respect, we
        # simply don't allow any name reuse across the entire model.
        self.name_resolution = parent.name_resolution if parent is not None else dict()
```

src/spox/_scope.py
```diff
@@ -114,18 +116,21 @@ def __delitem__(self, item: Union[str, H]):
...
         del self.name_of[value]
 
-    def enum(self, base: str, suffix: str = "_{}") -> str:
+    def enum(self, base: str) -> str:
         """Find an unused name by enumerating the pattern ``base + suffix.format(i)`` through `i = 0, 1, ...`"""
-        i = 0
-        while (name := f"{base}{suffix.format(i)}") in self:
-            i += 1
+        if base not in self.name_resolution:
+            self.name_resolution[base] = 0
```
cbourjau, line 122:
```suggestion
        self.name_resolution.setdefault(base, 0)
```

cbourjau (approved):
Thanks for this PR @aivanoved and thanks for the review @jbachurski ! I like this updated and simplified version much better! I just have a minor naming suggestion.

src/spox/_scope.py
aivanoved: Could be

CHANGELOG.rst
```diff
@@ -7,6 +7,14 @@
 Change log
 ==========
 
+(unreleased)
+-------------------
+
+**Other changes**
+
+- Imporved quadratic complexity of the naming algorithm in the build process, by keeping track of how many times a base name was used in current and all containing scopes.
```
cbourjau, line 15: The changelog is meant for users who don't necessarily care about the internals; only about how it affects them. We always update the unreleased version number with each PR according to Semver. Setting the correct version number at the time of the release can get necessarily annoying. 


```suggestion
0.12.1 (unreleased)
-------------------

**Other changes**

- The node-naming algorithm now has constant rather than quadratic time complexity.
```

cbourjau (approved):

CHANGELOG.rst
```diff
@@ -7,6 +7,14 @@
 Change log
 ==========
 
+0.12.1 (2024-XX-XX)
+-------------------
+
+**Other changes**
+
+- Imporved quadratic complexity of the naming algorithm in the build process, by keeping track of how many times a base name was used in current and all containing scopes.
```
JanJaguschQC, line 15:
```suggestion
- Improved quadratic complexity of the naming algorithm in the build process, by keeping track of how many times a base name was used in current and all containing scopes.
```

src/spox/_scope.py
```diff
@@ -114,18 +122,20 @@ def __delitem__(self, item: Union[str, H]):
         del self.of_name[key]
         del self.name_of[value]
 
-    def enum(self, base: str, suffix: str = "_{}") -> str:
+    def enum(self, base: str) -> str:
         """Find an unused name by enumerating the pattern ``base + suffix.format(i)`` through `i = 0, 1, ...`"""
```
jbachurski, line 126:
```suggestion
        """Find an unused name by enumerating the pattern ``f"{base}_{i}"`` through `i = 0, 1, ...`"""
```
This internal docstring is out of date now - something like this?

jbachurski (approved):
Looks better!
The behaviour is very slightly different now when a name would go unused, but it's probably imperceptible.

PR #160, opened 2024-06-18
keep the empty params remain empty

ReadyShowShow:
There are some operators that have optional parameters, just like Pad, which has the optional parameter constant_value, and like Resize, which has the optional parameters roi and scales.

It will edit empty parameters when there are empty parameters in the subgraph.

Therefore, I have fixed the bug.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,8 +7,13 @@
 Change log
 ==========
 
-0.12.1 (unreleased)
+0.12.1 (2024-06-18)
 -------------------
+
+**Bug fix**
+
+- Unset optional inputs are no longer erroneously prefixed by :func:`~spox.inline`.
+
 
 **Other changes**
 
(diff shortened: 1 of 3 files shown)


cbourjau:
@ReadyShowShow Thanks for opening this PR! I am having a little bit of trouble understanding what it does fix, though. Would you be able to add a test case or a post an example (in code) of the problem prior to this PR?

ReadyShowShow:
When you merge a subgraph, the names of all nodes and parameters will have a prefix before their original names. This also applies to empty parameters, and that's where the bug occurs.

ReadyShowShow:
just like this merged subgraph node:

NODE PROPERTIES
  type: Resize
  name: Inline_2__/Resize
INPUTS
  X: Inline_2__/Add_output_0
  roi: Inline_2__
  scales: Inline_2__
  sizes: Inline_2__/Concat_output_0

It will cause an error because the node 'Inline_2__' does not exist.

cbourjau:
Ok, I see the issue. I'll write a test case and push it here.

cbourjau (approved):

cbourjau:
Thank you for discovering this bug and providing the fix! I'll make a release today.

PR #164, opened 2024-07-22
Fix shape inference for Loop@v19 and v21

### LINKED ISSUE #163: Bugs in loop shape inference
### (v17) Custom inference fails when inputs and outputs of the body have mismatched shapes.

The following code should generate a loop that produces a tensor `[1 1 1 1 1]` by appending `1` in each iteration, but the inferred  shape is `(1,)`.
```Python
def test_loop():
    import spox.opset.ai.onnx.v17 as op
    import numpy as np
    
    num_iters = op.const(1)
    v = op.const([], dtype=np.int64)

    result = op.loop(num_iters, v_initial=[v], body=lambda i, c, x: (c, op.concat([x,op.const([1])], axis=0)))[0]
    assert result.type.shape == (5,)
```
Result:
```
FAILED bug.py::test_loop - assert (1,) == (5,)
```

### (v19+) Custom inference is missing completely

The fix from v17 is completely missing in all newer versions. The error can be reproduced using the following code
```Python
def test_loop2():
    import spox.opset.ai.onnx.v19 as op

    num_iters = op.const(1)
    v = op.const(1)

    result = op.loop(num_iters, v_initial=[v], body=lambda i, cond, x: (cond, x))[0]
    assert result.type.shape == ()
```
Result:
```
FAILED bug.py::test_loop2 - assert None == ()
```

Note: this test succeeds with v17

cbourjau:
Fixes #163 

# Checklist


src/spox/opset/ai/onnx/v21.py
```diff
@@ -612,6 +612,20 @@ class Inputs(BaseInputs):
...
+
+        body = self.attrs.body.value
+        n = len(body.requested_arguments) - 2
+
+        carried_names = list(self.outputs.get_vars())[:n]
+        carried_types = [v.type for v in list(body.requested_results.values())[1:][:n]]
+
+        for name, typ in zip(carried_names, carried_types):
+            output_types[name] = typ
+
```
adityagoel4512, line 626: Did you not want to check the output type of the body aligns with the initial value as per https://github.com/Quantco/spox/issues/163#issuecomment-2242788733?

adityagoel4512 (changes requested):

src/spox/opset/ai/onnx/v21.py
MatejUrbanQC: I'm not using loops in the end, so that's fine.

But I think this only make sense if you can explicitly mark dimensions that you want to change in the loop as `None`.

Let's say I want to use a function `lambda x: concat([x, [0]], axis=0)` in the loop.
If my input has shape `(None,)`, then the shape inference passes, because the output is also `(None,)`.
However, if my input has known shape, it will fail.

So it's not just very limiting, it's also inconsistent.

These issues should go away if you can say "This dimensions will be changed, treat it as `None`". If the number of dimensions will also change, you set the whole shape to `None`.

cbourjau:
I'm marking this as "draft" since the current implementation is too limiting / inconsistent.

adityagoel4512:
Will be superseded by https://github.com/Quantco/spox/pull/198.

PR #167, opened 2024-08-01
Run scheduled CI with onnx-weekly package

cbourjau:
Things seem to be in order, but better safe than sorry!


.github/workflows/weekly.yml
```diff
@@ -0,0 +1,62 @@
...
+        if: ${{ failure() && github.ref == 'refs/heads/main' }}
+        with:
+          script: |
+            github.rest.issues.listForRepo({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: "open",
+              labels: "[bot] onnx-weekly run"
+            }).then((issues) => {
+              if (issues.data.length === 0){
```
adityagoel4512, line 53: I'm not familiar with "github-script". Is `context.repo.owner` not the repo owner rather than the bot itself (which might be [`context.actor`](https://github.com/actions/toolkit/blob/main/packages/github/src/context.ts#L46)?

adityagoel4512 (approved):

.github/workflows/weekly.yml
cbourjau: As discussed elsewhere, let's simply try if this works as intended.

PR #172, opened 2024-08-07
Allow inline model to accept superset inputs (more inputs than expected)

### LINKED ISSUE #171: Inline sub-model cannot cope with superset inputs (more inputs than expected)
(as originally mentioned by @EmilyMohellenbrandQC).

https://github.com/Quantco/spox/blob/e88344b601d3dffd28de61af3bac96a05dc1a2a9/src/spox/_public.py#L292-L295

The current inline function checks if the supplied arguments `kwargs` are the exact same set as the inline model's expected inputs `in_names`. 

This could cause error if the supplied arguments are propagated down directly from the inputs of the main model, because the inputs to main model is likely to be a superset of the inputs to the sub-model.
e.g. If the main model is built with combining two inline sub-models 
```main_model = combine(inline(sub_model_a), inline(sub_model_b))```, 
where the `main_model` expects input `{"a", "b"}`, `sub_model_a` expects input `{"a"}` and `sub_model_b` expects input `{"b"}`. 
When the arguments `{"a", "b"}` are directly passed down to `sub_model_a`, this causes error.

A fix would be to check if `kwargs` is a superset of `in_names` instead of exact same set, i.e. 
`if not set(in_names).issubset(set(kwargs)):`

I tried the fix and it solves the issue.

EmilyMohellenbrandQC:
Currently inline models expect the supplied arguments to be the exact same set of its expected inputs.

Logically, as a sub-model of a bigger model, it should be able to handle superset input arguments that are passed down from main model.

This closes #171 

# Checklist

### DIFF
--- a/src/spox/_public.py
+++ b/src/spox/_public.py
@@ -289,7 +289,7 @@ def inline(model: onnx.ModelProto) -> _InlineCall:
                 array = array.astype(str)
             kwargs[name] = initializer(array)
 
-        if set(kwargs) != set(in_names):
+        if not set(in_names).issubset(set(kwargs)):
             raise TypeError(
                 f"Error processing arguments, got {set(kwargs)}, expected {set(in_names)}."
             )


EmilyMohellenbrandQC:
Decided to keep the current behavior of `inline` and accommodate the special cases by custom wrappers in downsream. See discussions in #171

PR #174, opened 2024-08-26
Add "const" to __all__

cbourjau:
This PR adds `"const"` to the default opsets' (`ai.onnx`) `__all__`. This will cause `const` to be correctly rendered in our documentation.

### DIFF
--- a/src/spox/opset/ai/onnx/v17.py
+++ b/src/spox/opset/ai/onnx/v17.py
@@ -15867,4 +15867,4 @@ _CONSTRUCTORS = {
     "Xor": xor,
 }
 
-__all__ = [fun.__name__ for fun in _CONSTRUCTORS.values()]
+__all__ = [fun.__name__ for fun in _CONSTRUCTORS.values()] + ["const"]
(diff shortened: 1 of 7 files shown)


src/spox/opset/ai/onnx/v17.py
```diff
@@ -15867,4 +15867,4 @@ def const(value: npt.ArrayLike, dtype: npt.DTypeLike = None) -> Var:
     "Xor": xor,
 }
 
-__all__ = [fun.__name__ for fun in _CONSTRUCTORS.values()]
+__all__ = [fun.__name__ for fun in _CONSTRUCTORS.values()] + ["const"]
```
adityagoel4512, line 15870: Is there a reason `_CONSTRUCTORS` only contains ONNX operators and not additions like `const`?

cbourjau: Frankly, I can't quite remember why we had the `_CONSTRUCTORS` in the first place. Either way, it seems like a mapping from class names to constructor functions, but `const` does not have a constructor class in the way the other operators do.

adityagoel4512 (approved):

PR #185, opened 2024-10-14
Enable pyupgrade

adityagoel4512:
This enables pyupgrade [again](https://github.com/Quantco/spox/blob/main/.pre-commit-config.yaml#L12-L17) but using ruff. It updates the various jinja templates accordingly.

# Checklist


cbourjau (approved):
ruff automagically updated the jinja2 templates?

adityagoel4512:
> ruff automagically updated the jinja2 templates?

I not so magically updated the jinja templates manually.

PR #189, opened 2024-11-06
Introduce `_VarInfo` internally to reduce memory footprint in value propagation

### LINKED ISSUE #187: Value propagation may produce a large memory footprint
To demonstrate the issue one may consult the following example:
```python
import tracemalloc
import spox
from spox.opset.ai.onnx import v20 as op
import numpy as np
import gc

def app():
    a = op.const(np.zeros(10000))
    b = op.const(np.zeros(10000))
    for i in range(N):
        a = op.mul(a, a)
        gc.collect()
    return a

tracemalloc.start()

var = app()

print(tracemalloc.get_traced_memory())

tracemalloc.stop()
```

Upon execution, the outputs are as follows:
```console
(spox) ➜  spox git:(main) ✗ python3 memtest10.py
(4494527, 4900843)
(spox) ➜  spox git:(main) ✗ python3 memtest100.py
(11821814, 12228285)
```

This example highlights that the `Var`-s persist with their values even when they are considered "unreachable" through the stable APIs. This issue stems from the current practice of consolidating all `Var` information in a singular linked [location](https://github.com/Quantco/spox/blob/main/src/spox/_var.py#L50-L53). Since most information is crucial during the subsequent building process, `Node`-s maintain references to their `Input` and `Output` entities, thus creating the present challenge.

Me and @cbourjau have been talking about ways to circumvent the issue at hand. In order of how hacky they are:
- Have the ownership mechanism be represented by some `weakref`-s. This would be extremely hacky to pull off, since we only need to garbage collect the computed values when all the paths are from the form `User`->`Var`->`Node`->...->`Var`->`_value`
- Have the user visible `Var`-s be a different object than the `Var` inside the creation graph. I am still a bit unsure whether this is something feasible as `spox` seems to rely on the fact that [copying](https://github.com/Quantco/spox/blob/main/src/spox/_var.py#L138-#L141) does not actually copy, however this invariant may be only needed for the building stage. 
- Create a `InnerVar` which holds all information except the `_value` which is held onto by every object in the graph and the current `Var` becomes a wrapper around that attaching a `_value` attribute. I am not extremely sure if this would yield a breaking change somewhere downstream.

This is the current stage of affairs, can you weight in? @jbachurski @adityagoel4512

neNasko1:
Fixes #187 by making the current user-facing `Var` be a wrapper around a helper class `VarInfo`, where `Var` is a pair `(VarInfo, PropValue)`.

The idea is to have `VarInfo` be the object that is passed around inside of all functions and be pointed to by all `Node`-s, so concrete propagated values can be garbage collected correctly. 

# Checklist for implementation

- [X] Provide better typing around `input_prop_values` dictionary
- [X] Fix `adapt_node`
- [X] Fix documentation 
- [X] Add a `CHANGELOG.rst` entry
- [X] Fix `Var`-s passed to manual inference
- [X] Make tests passing

# Breaking changes checklist
- [X] Examine the performance implications?
- [X] Examine downstream repos, does the change brake anything we rely on?


src/spox/opset/ai/onnx/ml/v3.py
```diff
@@ -662,13 +660,22 @@ def array_feature_extractor(
...
+    return (
+        _ArrayFeatureExtractor(
+            _ArrayFeatureExtractor.Attributes(),
+            _ArrayFeatureExtractor.Inputs(
+                X=unwrap_vars(X),
+                Y=unwrap_vars(Y),
+            ),
+            input_prop_values=input_prop_values,
+        )
+        .get_output_vars(input_prop_values=input_prop_values)
```
cbourjau, line 676: It is a bit surprising that we would need to supply `input_prop_values` twice here. Could we make it so that we only need to provide them to `get_output_vars`?

src/spox/_node.py
```diff
@@ -212,46 +206,40 @@ def pre_init(self, **kwargs):
...
+            else {}
+        )
 
-        for key, var in self.outputs.get_vars().items():
+        for key, var in self.outputs.get_var_infos().items():
             if var.type is None:  # If no existing type from init_output_vars
                 # Attempt to use the ones from kwargs, if none then what type inference gave
                 var.type = out_types.get(key)
 
+    def get_output_vars(self, input_prop_values={}):
```
cbourjau, line 239: I am surprised that no lint was complaining against using a mutable type as a default value. This should read `input_prop_values=None`.

src/spox/_node.py
```diff
@@ -130,7 +126,7 @@ def __init__(
             # As inference functions may access which output vars we initialized (e.g. variadics)
             # we inject uninitialized vars first
             self.outputs = self._init_output_vars()
-            self.inference(infer_types, propagate_values)
+            self.inference(infer_types, input_prop_values)
```
cbourjau, line 129: I would find it conceptually easier if we could separate the value propagation from the initialization of the `Node` object. Do you have an example where we may need the `input_prop_values` at this stage?

cbourjau (changes requested):
Thanks for taking this on and great work to already make the tests pass! I am trying to grok the details of the implementation. While doing so I already left a few high-level comments. I am not quite convinced that the use of a meta class is buying us much with respect to the introduced complexity/scariness of it. Would you mind adding type hints for its use and/or point out where it would provide a mypy-testable benefit compared to a simpler/dynamic structure such as a `dict[str, Var]`s?

jbachurski:
Are you sure there isn’t a cleaner way of doing this? It feels like a significant change. 
I wonder if your standpoint on value propagation changed, Christian @cbourjau - it was mostly intended as an experimental side-feature to improve experimentation back when I added it, not something that should be relied on in live systems. So I’m not sure a large change is worth it to improve the GC behaviour.
I’m afraid I won’t be able to help review this as I’m pretty starved on time.

src/spox/_exceptions.py
```diff
@@ -8,7 +8,7 @@
 
 
 class InferenceWarning(Warning):
-    """Warning related to partial typing of Variables.
+    """Warning related to partial typing of VarInfoiables.
```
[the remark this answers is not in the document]
neNasko1: done

src/spox/_node.py
```diff
@@ -212,46 +206,40 @@ def pre_init(self, **kwargs):
...
         """
         Inference routine for output types. Often overriden by inheriting Node types.
 
-        Returns a dictionary of output field names into Types for the respective Vars.
+        Returns a dictionary of output field names into Types for the respective VarInfos.
         """
         return {}
 
-    def inference(self, infer_types: bool = True, propagate_values: bool = True):
+    def inference(self, infer_types: bool = True, input_prop_values={}):
```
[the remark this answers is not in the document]
neNasko1: Sometimes inferring, one can infer some more type information by providing values in the graph. Examples for this is using `op.reshape`, `op.split`, `op.slice`.

src/spox/_fields.py
```diff
@@ -31,20 +34,68 @@ class VarFieldKind(enum.Enum):
...
+            else:
+                yield from ((f"{key}_{i}", v) for i, v in enumerate(value))
+
+    def get_var_infos(self) -> dict[str, Var]:
+        """Return a flat mapping by name of all the VarInfos in this object."""
+        return {key: var for key, var in self._flatten() if var is not None}
+
+
+class BaseVarsMeta(type):
+    def __new__(cls, name, bases, namespace):
```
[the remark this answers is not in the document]
neNasko1: removed

src/spox/opset/ai/onnx/ml/v3.py
neNasko1: done

src/spox/_var.py
```diff
@@ -24,28 +24,10 @@ def _not_impl(self, *_):
...
-    The ``type`` field is inferred and checked by operators.
-    It may be ``None`` if type inference failed, in which case it is unknown and should pass all type checks.
-    However, untyped ``VarInfo`` objects may not be used in some contexts.
-    Keep in mind that the types themselves may have some information missing.
-    For instance, tensors allow missing rank and shape information.
-
-    There is an implicit value propagation mechanism, powered by the ONNX reference implementation.
-    Values may be propagated if a ``VarInfo`` always has a known and constant value at runtime.
-    This is used for type & shape inference. For instance, Reshape to a constant shape can have the shape inferred.
+    Internal information about a ``Var``. Should be mainly inaccessible for most uses of ``spox``.
```
cbourjau, line 27: "most"? Does `VarInfo` become part of the public interface?

neNasko1: That is a good point, to be honest I was thinking about marking it as `_VarInfo` instead. I was really waiting to consolidate on the implementation to see if it had to pop-out somewhere.  

Now that I have gone through some of the downstream users of `spox`, I think it is safe to say we can mark it private.

src/spox/_standard.py
```diff
@@ -95,15 +99,31 @@ def out_value_info(curr_key, curr_var):
...
+            out_value_info(key, var)
+            for key, var in self.outputs.get_var_infos().items()
         ]
         # Initializers, passed in to allow partial data propagation
         #  - used so that operators like Reshape are aware of constant shapes
-        initializers = [
-            from_array(var._value.value, key)
-            for key, var in self.inputs.get_vars().items()
-            if var._value and isinstance(var._value.value, np.ndarray)
+        # TODO: fix this
```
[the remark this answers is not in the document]
neNasko1: I had to refactor a bit, good catch!

neNasko1:
> I'm afraid running this PR against [this branch](https://github.com/Quantco/ndonnx/tree/all-logic-in-data) in ndonnx (which makes extensive use of Spox's value propagation) gives rise to a lot of warnings of the type:
> 
> ```
> spox._exceptions.InferenceWarning:
>  Output type for variable expanded of ai.onnx@21::Unsqueeze was not concrete - ValueError: Tensor int32[...] does not specify the shape -- in ?.
> ```
> 
> These warnings do not appear when running on spox from the `main` branch.
> 
> You can make pytest fail on these warnings in the ndonnx test suite by running it as follows:
> 
> ```shell
> pytest tests -W="error:Output type for variable"
> ```
> 
> Do you think you can take a look into what is going on here?

This is caused by the fact that `validate_types` is being called before actually performing the value propagation. 

Currently this type-validation is ["optional"](https://github.com/Quantco/spox/blob/ca7ebbe9fb1dbfffdb8e3ffef1b2f7ac638e2180/src/spox/_node.py#L142), however AFAIK this is not really optional as `validate` is not exposed publicly. Should we remove this option. If we do, then we can safely move the validation after the value-propagation.

cbourjau (changes requested):
Thanks for looking into the warnings on the ndonnx side! I think this PR is converging :). Could you add type hints to all new functions please?

src/spox/_fields.py
```diff
@@ -31,20 +34,65 @@ class VarFieldKind(enum.Enum):
...
+        return self.vars[key]
+
+    def __setitem__(self, key: str, value) -> None:
+        """Allows dictionary-like access to set variables."""
+        self.vars[key] = value
+
+
 @dataclass
-class BaseVars(BaseFields):
+class BaseVarInfos(BaseFields):
```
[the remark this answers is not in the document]
neNasko1: I removed unnecessary methods and checked with `pytest` coverage to be sure that the remaining ones are needed.

CHANGELOG.rst
```diff
@@ -7,6 +7,13 @@
 Change log
 ==========
 
+0.14.0 (unreleased)
+-------------------
+
+**Other changes**
+
+- Split `Var`-s internally to help the garbage collector in collecting propagated values.
```
```suggestion
- Propagated values may now be garbage collected if their associated `Var` object goes out of scope. 
```

src/spox/_value_prop.py
```diff
@@ -24,9 +26,12 @@
 - PropValue -> Optional, Some (has value)
 - None -> Optional, Nothing (no value)
 """
-PropValueType = Union[np.ndarray, list["PropValue"], "PropValue", None]
-ORTValue = Union[np.ndarray, list, None]
-RefValue = Union[np.ndarray, list, float, None]
+PropValueType = Union[
```
cbourjau, line 29: Don't we get by using the explicit `list` type here? `Sequence` would otherwise be preferable over `Iterable`, too.

cbourjau (approved):
Very nice! I have one very tiny nit-pick and then this is ready! Thank you very much!

src/spox/_value_prop.py
neNasko1: Updated

PR #191, opened 2024-11-20
Remove logic around `Initializer`-s in `adapt_node`

neNasko1:
Currently it seems like `adapt_node` uses some of the constant propagated values as initialisers to the adapted node. 

After discussing this with @cbourjau it seems like it may be a legacy artefact as it does not break anything downstream. It may have existed in a way to ensure that the initialisers are properly put into the graph, however as demonstrated by the test case they get picked as attributes during the build procedure. 

# Checklist


cbourjau (approved):
Thanks!

PR #192, opened 2024-11-26
Deprecate operator dispatch in `Var`

neNasko1:
Operator dispatch is a too high-level feature for `spox`. Also it is currently clunky to use from a user standpoint as it essentially "monkey-patches" the dispatcher in `spox`.  

A more appropriate way to access this behaviour is by using [ndonnx](https://github.com/Quantco/ndonnx).

# Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,7 @@ Change log
 
 **Pending breaking changes**
 
-- Usage of operator overloading via ``_operator_dispatcher`` now triggers a ``DeprecationWarning``. This API is scheduled for removal in Spox ``0.14.0``. Consider using ``ndonnx`` as an alternative.
+- Usage of ``operator_overloading`` now triggers a ``DeprecationWarning``. This API is scheduled for removal in Spox ``0.14.0``. Consider using ``ndonnx`` as an alternative.
 
 **Other changes**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,7 @@ Change log
 
 **Pending breaking changes**
 
-- Usage of ``operator_overloading`` now triggers a ``DeprecationWarning``. This API is scheduled for removal in Spox ``0.14.0``. Consider using ``ndonnx`` as an alternative.
+- Importing :func:`spox._future.operator_overloading`` now triggers a ``DeprecationWarning``. The function will be removed in a future release. Consider using ``ndonnx`` as an alternative.
 
 **Other changes**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,10 @@ Change log
 
 - Value propagation of string tensors no longer raises an erroneous ``ValueError`` in some instances.
 
+**Pending breaking changes**
+
+- Usage of operator overloading via ``_operator_dispatcher`` now triggers a ``DeprecationWarning``. This API is scheduled for removal in Spox ``0.14.0``. Consider using ``ndonnx`` as an alternative.
+
 **Other changes**
 
 - The adaption logic is not using inferred values as initializers anymore.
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -18,6 +18,10 @@ Change log
 
 - Value propagation of string tensors no longer raises an erroneous ``ValueError`` in some instances.
 
+**Pending breaking changes**
+
+- Usage of ``operator_overloading`` now triggers a ``DeprecationWarning``. This API is scheduled for removal in Spox ``0.14.0``. Consider using ``ndonnx`` as an alternative.
```
cbourjau, line 23:
```suggestion
- Importing :func:`spox._future.operator_overloading`` now triggers a ``DeprecationWarning``. The function will be removed in a future release. Consider using ``ndonnx`` as an alternative.
```

cbourjau (changes requested):
I doesn't make a big difference, but technically we want to be sure that we warn users if they do anything that will break in the future. If we remove `operator_overloading` in the future any `from spox._future import operator_overloading` will already break regardless of usage. As such, we would ideally already warn on that level. One way to do so is to rename the function to `_operator_overloading` and then define a `getattr` function in the module to return it if a user imports "operator_overloading". We may then raise the warning from `getattr` rather than when the user actually calls the function. This comes with the added benefit that it would already remove the function from the documentation, too.

src/spox/_future.py
```diff
@@ -201,6 +202,10 @@ def operator_overloading(
     ...    return x * y
     >>> assert foo()._get_value() == np.array(6)
     """
+    warnings.warn(
+        "using operator_overloading is deprecated, consider using https://github.com/Quantco/ndonnx instead",
```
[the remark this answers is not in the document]
neNasko1: Done!

neNasko1:
> I doesn't make a big difference, but technically we want to be sure that we warn users if they do anything that will break in the future. If we remove `operator_overloading` in the future any `from spox._future import operator_overloading` will already break regardless of usage. As such, we would ideally already warn on that level. One way to do so is to rename the function to `_operator_overloading` and then define a `getattr` function in the module to return it if a user imports "operator_overloading". We may then raise the warning from `getattr` rather than when the user actually calls the function. This comes with the added benefit that it would already remove the function from the documentation, too.

I think it should now work as described here^

Note: These are how the different situations get handled after the change, using the [developer mode](https://docs.python.org/3/library/devmode.html#devmode):
```python
# warning
from spox._future import operator_overloading

# no warning
import spox._future as sf

# warning
sf.operator_overloading

# no warning
from spox._future import set_type_warning_level

# ImportError
from spox._future import non_existant_function
```

src/spox/_future.py
```diff
@@ -221,5 +233,6 @@ def operator_overloading(
     "set_value_prop_backend",
     "value_prop_backend",
     # Operator overloading on Var
-    "operator_overloading",
+    "operator_overloading",  # noqa: F822
+    "__getattr__",
```
cbourjau, line 237:
```suggestion
```
`__all__` is only used for IDEs and when generating the docs. It has not impact on whether something can actually be imported. We don't want to keep deprecated things around in the docs, so it is fine to remove it here. We should also remove it from the actual docs [here](https://spox.readthedocs.io/en/latest/manual/unstable.html#operator-overloading) in a follow-up PR.

cbourjau (approved):

src/spox/_future.py
neNasko1: > We should also remove it from the actual docs [here](https://spox.readthedocs.io/en/latest/manual/unstable.html#operator-overloading) in a follow-up PR.

Shouldn't this live in the PR removing the actual function?

CHANGELOG.rst
```diff
@@ -18,6 +18,10 @@ Change log
 
 - Value propagation of string tensors no longer raises an erroneous ``ValueError`` in some instances.
 
+**Pending breaking changes**
```
cbourjau, line 21:
```suggestion
**Deprecation**
```

PR #193, opened 2024-11-28
Remove unused `type:ignore`-s

neNasko1:
Upon reviewing the codebase, it appears that certain instances of `type: ignore` may not be necessary. As part of our aims to release `spox` 1.0, we should work towards strengthening our `mypy` setup to help with future maintenance.

I am open to hearing your comments. 

# Checklist


cbourjau (approved):
Thanks for doing this chore!

jbachurski:
For context, a bunch of these were related to shortcomings in PyCharm's checker, which I used developing Spox. So it should be fine to remove them.

PR #194, opened 2024-12-02
Improve type-hints in function definitions

neNasko1:
As there are current efforts to refactor `spox` and prepare it for a 1.0 release I have decided to look around the codebase for places where additional typing information can be provided. The changes are mainly around type-hinting in function declarations. 

This PR centres around the following diff and fixing the type-errors discovered by it:
```
diff --git a/pyproject.toml b/pyproject.toml
index 8f5e249..ec2459f 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ ignore_missing_imports = true
 no_implicit_optional = true
 check_untyped_defs = true
 warn_unused_ignores = true
+disallow_untyped_defs = true
 
 [tool.pytest.ini_options]
 # This will be pytest's future default.
```

`type: ignore`-s are only introduced in parts of the code that deals with the deprecated `operator_overloading` functionality.

### DIFF
--- a/src/spox/_schemas.py
+++ b/src/spox/_schemas.py
@@ -5,25 +5,21 @@
 
 import itertools
 from collections.abc import Iterable
-from typing import (
-    Callable,
-    Optional,
-    Protocol,
-    TypeVar,
-)
+from typing import Any, Callable, Optional, Protocol, TypeVar
 
 from onnx.defs import OpSchema, get_all_schemas_with_history
+
+
+class _Comparable(Protocol):
+    def __lt__(self, other: Any) -> bool: ...
+
+    def __gt__(self, other: Any) -> bool: ...
+
 
 S = TypeVar("S")
 K = TypeVar("K")
 V = TypeVar("V")
-T = TypeVar("T", bound="_Comparable", contravariant=True)
-
-
-class _Comparable(Protocol[T]):
-    def __lt__(self: T, other: T) -> bool: ...
-
-    def __gt__(self: T, other: T) -> bool: ...
+T = TypeVar("T", bound=_Comparable)
 
 
 def _key_groups(


src/spox/_function.py
```diff
@@ -89,11 +92,16 @@ def infer_output_types(self) -> dict[str, Type]:
...
+    def opset_req(self) -> set[tuple[str, int]]:
         node_opset_req = Node.opset_req.fget(self)  # type: ignore
         return node_opset_req | self.func_graph._get_build_result().opset_req
 
-    def update_metadata(self, opset_req, initializers, functions):
+    def update_metadata(
+        self,
+        opset_req: set[tuple[str, int]],
+        initializers: dict[Var, np.ndarray],
+        functions: list["Function"],
```
cbourjau, line 103: Does this work without the quotation marks  with
```python
from __future__ import annotations
```
?

This applies throughout and may trigger some [ruff reformatting](https://docs.astral.sh/ruff/rules/future-rewritable-type-annotation/).

src/spox/_adapt.py
```diff
@@ -80,7 +80,7 @@ def adapt_inline(
         base_model = node.model
         try:
             node.model = target_model
-            target_nodes = node.to_onnx(Scope.of((node, node_name), *var_names.items()))
+            target_nodes = node.to_onnx(Scope.of((node_name, node), *var_names.items()))
```
cbourjau, line 83: This looks a bit odd. How did this not raise an error earlier?

cbourjau (commented):
Thanks! I have just two minor comments.

src/spox/_adapt.py
neNasko1: You are correct, this is not needed. 
Removed

src/spox/_function.py
neNasko1: I am left under the impression that the `spox` codebase does not use `__future__` annotations. Do you suggest adding it in all relevant files? Should we leave it be for now and fix it in a different PR? 

Examples: [1](https://github.com/Quantco/spox/blob/main/src/spox/_build.py#L131), [2](https://github.com/Quantco/spox/blob/main/src/spox/_graph.py#L142), [3](https://github.com/Quantco/spox/blob/main/src/spox/_shape.py#L36)

cbourjau: We don't use it because we didn't know about it then (or at least I didn't). We should use `__future__`, though, since it makes the type hints so much nicer to write. This PR seems like a good place to do so. Could you add it to the relevant files, please?

jbachurski: I knew about `annotations`, but I remember something breaking. 
I think it may have been the places where type annotations are parsed (just operator definitions?).
You can try if it'll work fine now!

neNasko1: Yes the part in question that breaks is this [one](https://github.com/Quantco/spox/pull/194/files#diff-7aa15f9b066bdff1950605ed91fda2b81a1802f16cabc3ff64c2649764dc82eaR61-R72), where we parse the type hints. I think this fix ought to work.

src/spox/_fields.py
```diff
@@ -59,13 +59,15 @@ def __post_init__(self):
...
-        if field.type == Var:
+        # The field may be unannotated as per
+        # from __future__ import annotations
+        if field.type in [Var, "Var"]:
             return VarFieldKind.SINGLE
-        elif field.type == Optional[Var]:
+        elif field.type in [Optional[Var], "Optional[Var]"]:
             return VarFieldKind.OPTIONAL
-        elif field.type == Sequence[Var]:
+        elif field.type in [Sequence[Var], "Sequence[Var]"]:
```
cbourjau, line 70: I'm afraid this is not the ideal fix. We should instead use `typing.get_type_hints` to evaluate the hints properly. E.g.:

```python
typing.get_type_hints(cls)[field.name] == Optional[Var]
```

Note that this works:
```python
assert None | Var == Optional[Var]
```

src/spox/_function.py
cbourjau: Follow up: https://github.com/Quantco/spox/pull/194/files#r1869451563

src/spox/_schemas.py
```diff
@@ -14,20 +14,21 @@
...
 K = TypeVar("K")
 V = TypeVar("V")
-T = TypeVar("T", bound=_Comparable)
+T = TypeVar("T", bound="_Comparable", contravariant=True)
+
+
+class _Comparable(Protocol[T]):
+    def __lt__(self: T, other: T) -> bool: ...
+
+    def __gt__(self: T, other: T) -> bool: ...
```
cbourjau, line 26: This looks a little odd. The bound on `"_Comparable"` is implicitly a `"_Comparable[Any]"` anyhow. Let's just leave this largely as it was; only annotating `other: Any` and the return as `bool` without generics.

cbourjau (approved):

PR #197, opened 2024-12-16
Use `PEP604` type annotations

neNasko1:
As discussed with @cbourjau, this PR introduces the [PEP604](https://peps.python.org/pep-0604/) way of writing `Union`-s and `Optional`-s into the repo. 

This does not need a `CHANGELOG.rst` entry as it is not user-facing.


adityagoel4512 (approved):
Thanks!

PR #198, opened 2025-01-02
Fix shape inference for Loop

neNasko1:
This PR is based on the discussions in #163 and #164 and is a best effort solution to fill the type inferencing hole referenced. 

I provided a short [summary](https://github.com/Quantco/spox/issues/163#issuecomment-2568059691) of the current state to make it easier to follow the current discussion.

Note: This is not a complete implementation as mentioned by the comment in #163. 

# Checklist


src/spox/opset/ai/onnx/v17.py
```diff
@@ -1780,16 +1780,20 @@ class Outputs(BaseOutputs):
...
-        carried_names = list(self.outputs.get_var_infos())[:n]
-        carried_types = [v.type for v in list(body.requested_results.values())[1:][:n]]
+        initial_types = [v.type for v in list(body.requested_arguments)[2:]]
+        carried_types = [v.type for v in list(body.requested_results.values())[1:]]
 
-        for name, typ in zip(carried_names, carried_types):
-            output_types[name] = typ
+        is_constant_shape = all(
+            i_typ == c_typ for i_typ, c_typ in zip(initial_types, carried_types)
+        )
```
adityagoel4512, line 1793:
```suggestion
        shape_unchanged_between_iterations = all(
            i_typ == c_typ for i_typ, c_typ in zip(initial_types, carried_types)
        )
```

nitpick:  it's obvious enough from the context but "constant" is a bit of an overloaded term at this point.

tests/type_inference/test_loop.py
```diff
@@ -1,15 +1,36 @@
...
 # SPDX-License-Identifier: BSD-3-Clause
 
-import spox.opset.ai.onnx.v17 as op
+import numpy as np
+import pytest
+
+import spox.opset.ai.onnx.v17 as op17
+import spox.opset.ai.onnx.v19 as op19
+import spox.opset.ai.onnx.v21 as op21
 from spox import Tensor, argument
```
adityagoel4512, line 12: Please add tests with optionals and sequence initial value/outputs.

src/spox/opset/ai/onnx/v17.py
```diff
@@ -1780,16 +1780,20 @@ class Outputs(BaseOutputs):
...
+
+        for name, _, c_typ in zip(output_names, initial_types, carried_types):
+            output_types[name] = c_typ if is_constant_shape else c_typ._with_shape(None)
```
adityagoel4512, line 1796: Does `unsafe_reshape` + `unwrap_type` not do the job of `_with_shape`?

src/spox/opset/ai/onnx/v17.py
```diff
@@ -1780,16 +1780,20 @@ class Outputs(BaseOutputs):
...
+        output_types = super().infer_output_types(input_prop_values)
+        output_names = list(self.outputs.get_var_infos())
 
         body = self.attrs.body.value
-        n = len(body.requested_arguments) - 2
 
-        carried_names = list(self.outputs.get_var_infos())[:n]
-        carried_types = [v.type for v in list(body.requested_results.values())[1:][:n]]
+        initial_types = [v.type for v in list(body.requested_arguments)[2:]]
+        carried_types = [v.type for v in list(body.requested_results.values())[1:]]
```
adityagoel4512, line 1789: A very brief comment that `[2:]` skips the iteration number and condition and `[1:]` skips the returned loop condition will benefit future contributors.

adityagoel4512 (changes requested):
The overall approach looks good to me.

CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ Change log
 0.14.0 (unreleased)
 -------------------
 
+**Bug fix**
+
+- Apply custom shape inference logic in :func:`spox.opsets.ai.v19.loop` and :func:`spox.opsets.ai.v21.loop`.
+
 **Other changes**
 
 - Propagated values may now be garbage collected if their associated `Var` object goes out of scope.
```
adityagoel4512, line 19:
```suggestion
**Bug fix**

- Adds missing shape inference logic for :func:`spox.opsets.ai.v19.loop` and :func:`spox.opsets.ai.v21.loop`.

**Other changes**

- Propagated values may now be garbage collected if their associated `Var` object goes out of scope. 
- :func:`spox.opsets.ai.v17.loop`, :func:`spox.opsets.ai.v19.loop` and :func:`spox.opsets.ai.v21.loop` will only infer shapes for loop carried dependencies if their shapes are unchanged across iterations.
```

src/spox/opset/ai/onnx/v17.py
neNasko1: AFAICS, this does not do the job as `unsafe_reshape` works on a `Var`, moreover seems like it works on `Tensor` shapes only.

src/spox/opset/ai/onnx/v17.py
neNasko1: Thanks for the suggestions!

src/spox/_type_system.py
```diff
@@ -56,6 +56,12 @@ def _from_onnx(cls, proto: onnx.TypeProto) -> Type:
             f"Cannot get Type from invalid protobuf (not tensor, sequence or optional): {proto}"
         )
 
+    def _with_shape(self, new_shape: SimpleShape) -> Type:
+        """
+        Creates a copy of the Type, but with a changed shape property
+        """
+        return self
```
adityagoel4512, line 63: I'd much rather this was an abstract method and we got an error than have this default behavior.
You override this for all subclasses, but this will immediately require work if we some day introduce `Map` to Spox.

I also think this is too generic: based on https://github.com/onnx/onnx/blob/main/onnx/onnx.proto3#L727, you can technically have arbitrarily nested `Sequence`s. For `loop` shape inference, we only allow one level depth (`Sequence(Tensor)`, see https://github.com/onnx/onnx/blob/main/docs/Operators.md#type-constraints-88). I think a distinct free function that simply deals with the nesting we care about for `loop` shape inference is a better option than a generic member on `Type`. Furthermore, everything you need is already in the public API of the `Type` variants.

TLDR: please make a free function that only accepts the typing arrangements in https://github.com/onnx/onnx/blob/main/docs/Operators.md#type-constraints-88.

neNasko1:
Note, that this does not run the test actions, so we should wait before merging.

src/spox/_type_inference_utils.py
```diff
@@ -0,0 +1,26 @@
...
+            raise InferenceError(
+                f"Type {typ} not allowed for state variables in Loop operator, sequence element can only be a tensor"
+            )
+        return Sequence(loop_erase_shape_info(typ.elem_type))
+    elif isinstance(typ, Optional):
+        if isinstance(typ.elem_type, Optional):
+            raise InferenceError(
+                f"Type {typ} not allowed for state variables in Loop operator, optionals of optionals are not allowed"
+            )
+        return Optional(loop_erase_shape_info(typ.elem_type))
```
adityagoel4512, line 23:
```suggestion
        return Sequence(Tensor(typ.elem_type, None))
    elif isinstance(typ, Optional):
        if not isinstance(typ.elem_type, Sequence | Tensor):
            raise InferenceError(
                f"Type {typ} not allowed for state variables in Loop operator, optionals of optionals are not allowed"
            )
        return Optional(loop_erase_shape_info(typ.elem_type))
```

adityagoel4512: Again, I think it's better to be completely explicit so this works just as well when `Type` expands. Otherwise, this is good to go, thanks!

adityagoel4512 (dismissed):
Thanks, let's fix test CI then merge!

CHANGELOG.rst
```diff
@@ -10,9 +10,14 @@ Change log
...
 -------------------
 
+**Bug fix**
+
+- Adds missing shape inference logic for :func:`spox.opsets.ai.v19.loop` and :func:`spox.opsets.ai.v21.loop`.
+
 **Other changes**
 
 - Propagated values may now be garbage collected if their associated `Var` object goes out of scope. 
+- :func:`spox.opsets.ai.v17.loop`, :func:`spox.opsets.ai.v19.loop` and :func:`spox.opsets.ai.v21.loop` will only infer shapes for loop carried dependencies if their shapes are unchanged across iterations.
```
adityagoel4512, line 21:
```suggestion


```

adityagoel4512 (approved):

neNasko1:
Thanks for the help 🚀

PR #199, opened 2025-01-02
Remove unused `jinja` templates

neNasko1:
This PR removes some of the old/unused `jinja` templates. 

Also based on the following upstreamed changes we can safely remove some of the custom type propagation subroutines:
- [Binarizer](https://github.com/onnx/onnx/pull/4881)
- [Compress](https://github.com/onnx/onnx/pull/4812)
- [OneHotEncoder](https://github.com/onnx/onnx/pull/4813)

Since we do some extra checking around the following operators, I am asking for input whether to include them in this PR. 
- [CategoryMapper](https://github.com/onnx/onnx/pull/4600)
- [TreeEnsembleRegressor](https://github.com/onnx/onnx/pull/4600)
- [TreeEnsembleClassifier](https://github.com/onnx/onnx/pull/4600)

EDIT: We [decided](https://github.com/Quantco/spox/pull/199#discussion_r1904584292) to wait for those 3 operators until some later `onnx` release. 

# Checklist


src/spox/opset/ai/onnx/ml/v3.py
```diff
@@ -124,18 +121,6 @@ class Inputs(BaseInputs):
...
-        if not self.inputs.fully_typed:
-            return {}
-        cats1, cats2 = self.attrs.cats_int64s, self.attrs.cats_strings
-        if cats1 is None or cats2 is None:
-            raise InferenceError("Missing required attributes.")
-        if len(cats1.value) != len(cats2.value):
-            raise InferenceError("Categories lists have mismatched lengths.")
-        t = self.inputs.X.unwrap_tensor()
-        (elem_type,) = {np.int64, np.str_} - {t.dtype.type}
-        return {"Y": Tensor(elem_type, t.shape)}
```
adityagoel4512, line 137: It looks like the only difference is that attributes are not being validated for existence and length in https://github.com/onnx/onnx/pull/4600/files#diff-cf90e45168824f70d9af1f90cd777a4542f2637bb4f3871542679c1a0a204b82R168-R179, unlike here. Can you please upstream attribute validation before removing the patch here. See the TreeEnsemble shape inference code for an example of the utilities: https://github.com/onnx/onnx/blob/main/onnx/defs/traditionalml/defs.cc#L1137-L1152.

src/spox/opset/ai/onnx/ml/v3.py
```diff
@@ -375,20 +360,6 @@ class Inputs(BaseInputs):
...
-            n_encodings = len(self.attrs.cats_int64s.value)
-        elif self.attrs.cats_strings:
-            n_encodings = len(self.attrs.cats_strings.value)
-        else:
-            raise InferenceError(
-                "Either `cats_int64s` or `cats_strings` attributes must be set."
-            )
-        shape = (*self.inputs.X.unwrap_tensor().shape, n_encodings)  # type: ignore
-        return {"Y": Tensor(dtype=np.float32, shape=shape)}
-
```
adityagoel4512, line 391: I believe the exact same comment as for CategoryMapper applies here, except that we already check for existence (just not length) upstream. After that we can remove this.

src/spox/opset/ai/onnx/ml/v3.py
```diff
@@ -528,30 +499,6 @@ class Outputs(BaseOutputs):
...
-            )
-        if self.inputs.fully_typed:
-            shape = self.inputs.X.unwrap_tensor().shape
-            assert shape is not None  # already checked with fully_typed
-            if len(shape) != 2:
-                raise InferenceError("Expected input to be a matrix.")
-            n = shape[0]
-        else:
-            n = None
-        return {"Y": Tensor(y_type, (n,)), "Z": Tensor(np.float32, (n, e))}
```
adityagoel4512, line 553: This looks good to remove to me 👍. Technically we should have validated the attributes here or upstream, but this doesn't remove any existing functionality.

src/spox/opset/ai/onnx/ml/v3.py
```diff
@@ -592,19 +539,6 @@ class Inputs(BaseInputs):
...
-            assert shape is not None  # already checked with fully_typed
-            if len(shape) != 2:
-                raise InferenceError("Expected input to be a matrix.")
-            assert shape is not None
-            n = shape[0]
-        else:
-            n = None
-        e = self.attrs.n_targets.value if self.attrs.n_targets is not None else None
-        return {"Y": Tensor(np.float32, (n, e))}
-
```
adityagoel4512, line 607: Looks good to remove.

src/spox/opset/ai/onnx/v17.py
```diff
@@ -499,32 +499,6 @@ class Inputs(BaseInputs):
...
-            axis = self.attrs.axis.value
-            if not (-len(shape) <= axis < len(shape)):
-                raise InferenceError(
-                    f"Compress attribute 'axis' must in range [-rank, rank-1] (rank={len(shape)})."
-                )
-            shape[axis] = None
-        else:
-            shape = [None]
-        return {"output": Tensor(inp.dtype, tuple(shape))}
-
```
adityagoel4512, line 527: Looks good to remove.

tests/type_inference/test_tree_ensemble_classifier.py
```diff
@@ -11,7 +11,7 @@
 def test_tree_ensemble_regressor_inference_str():
     (x,) = arguments(x=Tensor(np.float32, ("N", 5)))
     y, z = op_ml.tree_ensemble_classifier(
-        x, classlabels_strings=["a", "b"], class_ids=[1, 2, 3]
+        x, classlabels_strings=["a", "b", "c"], class_ids=[1, 2, 3]
```
adityagoel4512, line 14: Can you explain why you made this change?

tools/templates/extras/xloop.jinja2
```diff
@@ -1,33 +0,0 @@
-def xloop(
```
adityagoel4512, line 1: It looks like `xloop` was made private at some point and then removed. Let's let @jbachurski or @cbourjau weigh in briefly for some historical context first, but to my eye this looks fine.

adityagoel4512 (changes requested):

src/spox/opset/ai/onnx/ml/v3.py
neNasko1: Let me try to get all the diffs into upstream `onnx`, should not be that hard, and make the merge contingent on not losing  any functionality.

tests/type_inference/test_tree_ensemble_classifier.py
neNasko1: Good catch. 

I should have described this change in the PR. I think there was a bug in this test as [onnx](https://onnx.ai/onnx/operators/onnx_aionnxml_TreeEnsembleClassifier.html) states:
> One and only one of classlabels_strings or classlabels_int64s will be defined. The class_ids are indices into this list. 

Before this change, the upstream type propagation was failing.

tools/templates/extras/xloop.jinja2
jbachurski: I don’t remember this very well, so it may have been an attempt to make the API more convenient. It’s probably fine to remove since it’s unused.

src/spox/opset/ai/onnx/ml/v3.py
neNasko1: I think it should be done!

adityagoel4512: Thanks! Let's wait for a `onnx` release (or two) before removing the patch though.

neNasko1: Should be done now, can you take a look?

adityagoel4512 (approved):

PR #201, opened 2025-01-15
Introduce internal interface for `ValuePropBackend`

neNasko1:
Since we currently "bless" the reference and onnxruntime implementations as the only way to propagate value internally, this PR refactors a bit the internals to allow a better interface for "custom" execution providers(i.e. @cbourjau rust runtime, onnx-mlir...). 
Consult the [tests](https://github.com/Quantco/spox/pull/201/files#diff-4fd3c1127443dc7813c4d2f5f6be9428b706028bf55bd72e11132635de73394fR239) for a POC which could also be used to register custom opsets as well.

An additional internal change is related to the `PropDict` type as previously it was too wide and didn't capture the correct semantics of the returned values from onnxruntime. As per the definition of [from_ref_value](https://github.com/Quantco/spox/blob/386bccb6e769c4f228f93f5bb29871bb91df875a/src/spox/_value_prop.py#L106) we can see that the returned feed is of type `dict[str, PropValue]` which makes more sense to model around. 

This exposed a bug in the function value inference which was unnoticed and is now fixed.  

# Checklist


src/spox/_function.py
```diff
@@ -86,10 +86,12 @@ def infer_output_types(self, input_prop_values: PropDict) -> dict[str, Type]:
 
         self.func_inputs = self.Inputs(**self.func_args)
         self.func_outputs = self.constructor(
-            self.func_attrs, self.func_inputs.into_vars(input_prop_values)
+            self.func_attrs,
+            self.func_inputs.into_vars(input_prop_values),  # type: ignore
         )
+        # TODO: Implement function propagation
```
adityagoel4512, line 92: Is this PR ready or is the TODO superfluous?

neNasko1: This TODO reflect a previous bug that got introduced in #189. It was uncovered after restricting the `PropDict` to be `dict[str, PropValue]` instead of `dict[str, PropValueType]` and has to do with the `into_vars` which gets called on the output `_VarInfos`-s with the input propagated values.

adityagoel4512: There are quite a lot of moving parts in this PR. Is this change absolutely necessary to introducing the `BaseValuePropBackend`? It would be helpful to have one PR for fixing that bug and then a PR that just handles the value propagation interface.

src/spox/_value_prop_backend.py
```diff
@@ -0,0 +1,123 @@
...
+from typing import TYPE_CHECKING, Any, Generic, TypeVar
+
+import onnx
+
+from . import _value_prop
+from ._type_system import Type
+from ._value_prop import ORTValue, PropValue, RefValue
+
+if TYPE_CHECKING:
+    pass
```
adityagoel4512, line 18:
```suggestion
```

src/spox/_value_prop_backend.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+            return {}
+        return output_feed
+
+    def unwrap_feed(self, typ: Type, value: RefValue) -> PropValue:
+        return PropValue.from_ref_value(typ, value)
+
+
+class OnnxruntimeValuePropBackend(BaseValuePropBackend[ORTValue]):
+    def __init__(self, session_options: dict[str, Any] | None = None) -> None:
```
adityagoel4512, line 63: I don't think we need the `session_options` field here. It isn't used within Spox and as for custom operators, the ABC is sufficient.

src/spox/_value_prop_backend.py
```diff
@@ -0,0 +1,123 @@
...
+        self.session_options = onnxruntime.SessionOptions()
+
+        if session_options is None:
+            session_options = {"log_severity_level": 3}
+
+        for opt, val in session_options.items():
+            setattr(self.session_options, opt, val)
+
+    def wrap_feed(self, value: PropValue) -> ORTValue:
+        return value.to_ort_value()
```
adityagoel4512, line 75: It's not clear to me why you are leaving the `PropValue` -> `ORTValue` conversion in the `PropValue`. This is the only conceivable call site right? Same applies for `RefValue` and the opposite translation.

src/spox/_value_prop_backend.py
```diff
@@ -0,0 +1,123 @@
...
+        return PropValue.from_ref_value(typ, value)
+
+
+_VALUE_PROP_BACKEND: BaseValuePropBackend | None = ReferenceValuePropBackend()
+
+
+def get_value_prop_backend() -> BaseValuePropBackend | None:
+    if _VALUE_PROP_BACKEND is not None:
+        return _VALUE_PROP_BACKEND
+    elif _value_prop._VALUE_PROP_BACKEND == _value_prop._VALUE_PROP_BACKEND.REFERENCE:
```
adityagoel4512, line 106: Why are you maintaining both `_value_prop._VALUE_PROP_BACKEND` AND `_value_prop_backend._VALUE_PROP_BACKEND`? We should not need the former any more.

src/spox/_var.py
```diff
@@ -374,4 +374,8 @@ def create_prop_dict(
 
     flattened_vars = BaseVars(kwargs).flatten_vars()
 
-    return {key: var._value for key, var in flattened_vars.items() if var is not None}
+    return {
+        key: var._value
+        for key, var in flattened_vars.items()
+        if var is not None and var._value is not None
```
adityagoel4512, line 380: This looks related to https://github.com/Quantco/spox/pull/201/files#r1918654725. I'd like to discuss that in a separate PR if that can be disentangled out.

adityagoel4512 (changes requested):

src/spox/_value_prop_backend.py
neNasko1: I think we discussed keeping the old API and not introduce a breaking change still. This PR introduces a deprecation warning when using the old API, but it should still work.

src/spox/_var.py
neNasko1: Will split them out!

src/spox/_value_prop_backend.py
adityagoel4512: It would be nice to not break `spox._future.set_value_prop_backend` suddenly (therefore we should warn our users). The `spox._value_prop._VALUE_PROP_BACKEND` field on the other hand is just a private global variable.

src/spox/_future.py
```diff
@@ -35,19 +37,46 @@ def type_warning_level(level: TypeWarningLevel) -> Iterator[None]:
...
+            DeprecationWarning,
+        )
+
+    new_backend: spox._value_prop_backend.BaseValuePropBackend | None = None
+    if backend == spox._value_prop.ValuePropBackend.REFERENCE:
+        new_backend = spox._value_prop_backend.ReferenceValuePropBackend()
+    elif backend == spox._value_prop.ValuePropBackend.ONNXRUNTIME:
+        new_backend = spox._value_prop_backend.OnnxruntimeValuePropBackend()
+    elif isinstance(backend, BaseValuePropBackend):
+        new_backend = backend
```
adityagoel4512, line 76: Why are you not using `set_value_prop_backend` here?

CHANGELOG.rst
```diff
@@ -14,10 +14,15 @@ Change log
 
 - Adds missing shape inference logic for :func:`spox.opsets.ai.v19.loop` and :func:`spox.opsets.ai.v21.loop`.
 
+**Deprecation**
+
+- Using :func:`spox._future.set_value_prop_backend`` with a `spox._future.ValuePropBackend`` now triggers a ``DeprecationWarning``.
```
adityagoel4512, line 19: Are you proposing that each time a user wants to do value propagation, they extend `BaseValuePropBackend`? While the additional flexibility is valuable (for example with custom operators where you want to link your custom operator library), does it now become quite verbose simply to enable value propagation with vanilla ORT?

src/spox/_value_prop_backend.py
neNasko1: Sorry, the mistake is mine I thought it was a different diff. I agree

CHANGELOG.rst
neNasko1: I want to make public `OnnxruntimeValuePropBackend` and `ReferenceValuePropBackend`. Maybe it got left out, will add it back.

src/spox/_future.py
neNasko1: This is currently a minimum effort way to keep the old interface working for the time being. In this case it doesn't make sense to use `set_value_prop_backend` since this is a context manager and it's better to lend the work to the more suited [new](https://github.com/Quantco/spox/pull/201/files#diff-00fd4864cf41f57d773f9631cf1aa32c04bc0e99f711875ad64c6c09b03f8000R118-R123) context manager.

src/spox/_value_prop_backend.py
neNasko1: Since the `PropValue`-s are generally accessible from the public API and those are public functions of them I guess this is not the only call site. I am not sure if this is true in practice, i.e. I don't think any users are using that functionality.

The better point in keeping them this way is that the `to_ref_value` and `to_ort_value` are [intertwined](https://github.com/Quantco/spox/blob/main/src/spox/_value_prop.py#L141C1-L159C30) so it doesn't make sense to split them apart.

neNasko1:
@adityagoel4512 I have left the open conversations here so you could review them on #202 
Sorry for the inconvinience!

PR #202, opened 2025-01-21
Introduce internal interface for `ValuePropBackend`

neNasko1:
Duplicate of #201 because of missing access to the repo

Since we currently "bless" the reference and onnxruntime implementations as the only way to propagate value internally, this PR refactors a bit the internals to allow a better interface for "custom" execution providers(i.e. @cbourjau rust runtime, onnx-mlir...). 
Consult the [tests](https://github.com/Quantco/spox/pull/201/files#diff-4fd3c1127443dc7813c4d2f5f6be9428b706028bf55bd72e11132635de73394fR239) for a POC which could also be used to register custom opsets as well.

# Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,10 @@ Change log
 
 - Adds missing shape inference logic for :func:`spox.opsets.ai.v19.loop` and :func:`spox.opsets.ai.v21.loop`.
 
+**Deprecation**
+
+- Using :func:`spox._future.set_value_prop_backend`` now triggers a ``DeprecationWarning``. The function will be removed in a future release.
+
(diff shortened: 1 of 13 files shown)


cbourjau:
Apologies for letting this PR go stale! As mentioned in conversations back in January, I think we should strive to find a more minimalistic API. Conceptually, all we need is a function `(ModelProto, dict[str, np.ndarray]) -> dict[str, PropValue]`. Better yet, we should even consider a function `(ModelProto) -> dict[str, PropValue]` if we bake the inputs into the model as constants. 

I'll close this PR for now, but I hope that we'll revive it with such an API in the near future.

PR #228, opened 2025-06-05
Always use an explicit IR version when calling make_model

cbourjau:
`onnx.helpers.make_model` defaults to the latest IR version if unset. Runtimes can only support IRs up to a version known at their release. This default behavior is not news to us. When building a model using `spox.build`, we consistently set the IR version to the very conservative value 8 ~~(now bumped to 10 in this PR)~~ already. However, when building models for value propagation we didn't do the same and thus got the very latest IR version 11. This creates issues if we want to propagate values in environments with older onnxruntimes installed. This PR ensures that we also build models in those scenarios using a consistent IR and thus have a working value propagation.

# Checklist


CHANGELOG.rst
```diff
@@ -7,6 +7,13 @@
 Change log
 ==========
 
+0.15.0 (2025-06-05)
+-------------------
+
+**Bug fix**
+
+- All models (including those used for value propagation operations) are now consistently built using IR version 10 introduced in ONNX 1.16.0. Previously, models built using :func:`spox.build` used IR version 8.
+
```
adityagoel4512, line 16: Isn't this two different changes? Firstly, bumping IR version and secondly (the bug fix), setting the IR version for value propagation.

src/spox/_utils.py
```diff
@@ -64,3 +65,27 @@ def from_array(arr: np.ndarray, name: str | None = None) -> TensorProto:
...
+
+def make_model(
+    graph: onnx.GraphProto,
+    *,
+    opset_imports: list[onnx.OperatorSetIdProto] | None = None,
+    producer_name: str | None = None,
+    doc_string: str | None = None,
+    functions: list[onnx.FunctionProto] | None = None,
+) -> onnx.ModelProto:
+    """Like ``onnx.helper.make_model`` but with a consistent and reasonable default IR version."""
```
adityagoel4512, line 78:
```suggestion
    """Like ``onnx.helper.make_model`` but with a default IR version that is runnable on onnxruntime."""
```

adityagoel4512 (approved):
It would be good to add an older onnxruntime environment which we can run in CI to catch this in the future.

CHANGELOG.rst
cbourjau: Fair point! I'll roll back the bump to 10 as it does not give us anything.

PR #235, opened 2025-06-18
Use raw encoding for numerical tensors

cbourjau:
Protobuf tensor creation can be annoyingly slow for large tensors. This is particularly an issue for deep learning use cases and artifacts with large lookup tables. This change results in speedups of 10x or more (for the conversion time; runtime is unaffected). 

Running the following benchmark script:
```python
from time import perf_counter
import numpy as np

import spox.opset.ai.onnx.v21 as op

for size_exp in range(1, 9):
    size = 10**size_exp

    arr = np.ones((size,), dtype=np.float32)
    t0 = perf_counter()
    for _ in range(5):
        op.const(arr)
    t1 = perf_counter()

    print(f"{size=}: {(t1 - t0)}s")
```

yields the following results prior to this PR ("without raw") and after this PR:
```
# Without raw:

size=10: 0.003982624999480322s
size=100: 0.0008588329947087914s
size=1000: 0.0018544170015957206s
size=10000: 0.011751874990295619s
size=100000: 0.10788500000489876s
size=1000000: 1.087528750009369s
size=10000000: 11.295649749983568s
size=100000000: 122.66386904200772s

# With raw:

python bench.py
size=10: 0.001316792011493817s
size=100: 0.0008856659987941384s
size=1000: 0.0007787499926052988s
size=10000: 0.0015762920083943754s
size=100000: 0.0034534579899627715s
size=1000000: 0.0508717920165509s
size=10000000: 0.6581135420128703s
size=100000000: 12.970771457999945s
```

# Checklist


CHANGELOG.rst
```diff
@@ -7,6 +7,14 @@
 Change log
 ==========
 
+0.16.0 (2025-06-19)
```
adityagoel4512, line 10: Is there a reason you'd like to release this today?

tests/test_value_propagation.py
```diff
@@ -245,3 +245,16 @@ def test_strings():
 
     x, y = op.const(["foo"]), op.const(["bar"])
     np.testing.assert_equal(op.string_concat(x, y)._value.value, np.array(["foobar"]))  # type: ignore
+
+
+@pytest.mark.parametrize("dtype", [np.int32, np.float32])
+def test_tensor_raw_bytes(dtype):
+    # - Test on integers and floats to check for possible issues with
+    # endianess.
```
adityagoel4512, line 253: I don't see how this checks any "endianess" issues? You are turning these into ONNX tensors and recovering them back on the same platform. It also doesn't look like something that belongs as a Spox unit test but rather the upstream onnx library, if it isn't already tested.

src/spox/_utils.py
```diff
@@ -50,8 +50,20 @@ def from_array(arr: np.ndarray, name: str | None = None) -> onnx.TensorProto:
...
+    # Using raw=True does not allow for meaningful inspection of the
+    # values in netron and elsewhere. However, creating large protobuf
+    # tensors with `raw=False` is an order of magnitude slower for
+    # large tensors. The following makes the compromise that large
+    # numerical arrays can be encoded in the raw format since they
+    # defy manual inspection anyhow. Creating the tensor is still
+    # rather involved and also requires handling of endianess. We thus
+    # still fall back to using `onnx.numpy_helper.from_array`, which
+    # uses raw=True under the hood.
+    if arr.size > 100 and arr.dtype not in [np.str_, np.object_]:
```
adityagoel4512, line 65: I think it's a little strange to base this on netron's capabilities. Can we make a PR there to make raw bytes rendered? It looks like data is stored in little endian format in protobuf and the data type is known, so I imagine netron _could_ support this.

(see https://github.com/onnx/onnx/blob/main/onnx/numpy_helper.py#L530-L532 for the write path and then https://github.com/onnx/onnx/blob/main/onnx/numpy_helper.py#L370-L373 for the read path)

adityagoel4512 (changes requested):

tests/test_value_propagation.py
cbourjau: The data still goes through onnxruntime. Nonetheless, you are right that this does not fall into Spox's responsibility, and things appear to work as expected anyhow. I removed the test.

cbourjau:
Prompted by your [comment](https://github.com/Quantco/spox/pull/235#discussion_r2156805200), I went ahead and tried things out (thanks!). Everything works as expected in Netron already! I dug into the archives to figure out why we had the current workaround in place. It's from the old, closed-source days and boils down to the readability/editability of the protobuf itself rather than anything Netron-related. That is not a good enough reason to keep the observed performance penalty, IMHO. I removed the override completely in favor of just using the upstream function.

CHANGELOG.rst
```diff
@@ -7,6 +7,14 @@
...
+-------------------
+
+**New feature**
+
+- The instantiation time of constant numerical tensors with more than 100 elements has been significantly reduced by encoding their payload as raw bytes.
```
```suggestion
- The instantiation time of constant numerical tensors has been significantly reduced by encoding their payload as raw bytes.
```

CHANGELOG.rst
cbourjau, line 10:
```suggestion
0.16.0 (2025-06-20)
```

Issue #28, opened 2023-02-09
Tracking issue for unsupported parts of the standard


JakubBachurskiQC:
There are some parts of the standard that we don't yet support or we can't test due to their limited usage, but it would be a good idea to write these down in case they come up.

Unticked features written in **bold** have (unstable and possibly partial) internal implementations and test suites. Let ticked features be the ones that are stable. 

Features in this list without their own issue are **not significant enough to be a blocker** for a release.

- [ ] [**Functions**](https://github.com/onnx/onnx/blob/main/docs/IR.md#functions) - Limited testing due to limited ORT support
- [ ] [**Initialisers**](https://github.com/onnx/onnx/blob/main/docs/IR.md#graphs) - Replaceable by `Constant`, initialisers as input defaults unsupported in ORT. #50
- [ ] [External tensor data](https://github.com/onnx/onnx/blob/main/docs/IR.md#external-tensor-data) - Currently values that become `TensorProto` must be kept in memory as `numpy.ndarray`. How will this affect running type inference, when e.g. attributes are expected?
- [ ] Sparse tensors - Mostly seem to have a use in initialisers, but currently their appearances for e.g. `Constant` are ignored.
- [ ] Map value type - Though in the standard and undoubtedly would be useful, I'm not sure if it's possible to create/use one.
- [x] Non-standard (non-numpy) data types - though unmentioned (?) in the standard directly, some operators accept dtypes like `bfloat16`. We would need to slightly modify our representation to accept this.
- [ ] Custom naming - Note this isn't strictly required by the standard for producers but it sometimes comes up (e.g. for partial output evaluation). Since Spox tries to keep errors from reaching ONNX checkers it autogenerates all names. It could be useful to allow the user to use their own naming, or access Spox's, if they really wish to.
- [ ] [_Training, differentiability_](https://github.com/onnx/onnx/blob/main/docs/IR.md#training-related-information) - This seems to be mostly a preview feature.

jbachurski:
I can update that I did in fact encounter a Map in the wild, see discussion [here](https://github.com/onnx/onnx/issues/5006#issuecomment-1530360359). skl2onnx occasionally generates them, even though they are essentially useless, as shape inference in ONNX used to crash whenever they were encountered, which I fixed [here](https://github.com/onnx/onnx/pull/5188).
In any case, it might be time to implement them soon.

jbachurski:
Some thoughts on missing features for future consideration:

[...]

PR #265, opened 2025-12-22
Support all data types used in version 22 of the ai.onnx opset

cbourjau:
The onnx standard has more data types than those officially supported by spox today. The `onnx` package has supported these data types through the `ml_dtypes` package for a while. Since we pass through the dtype object to `onnx` at the relevant places, we also gain support for these data types automatically in spox.

This PR makes that support official and adds relevant tests to avoid regressions. Once we get the latest onnxruntime on conda-forge we can add the data types that were introduced with ai.onnx >=23.

# Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,13 +25,13 @@ repos:
       # ruff
       - id: ruff
         name: ruff
-        entry: pixi run -e lint ruff check --fix --exit-non-zero-on-fix --force-exclude --target-version py39
+        entry: pixi run -e lint ruff check --fix --exit-non-zero-on-fix --force-exclude
         language: system
         types_or: [python, pyi]
         require_serial: true
       - id: ruff-format
         name: ruff-format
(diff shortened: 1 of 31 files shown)


jbachurski:
I just wanted to chime in that this is a perfectly reasonable idea (and we get to tick a thing off #28).
May I suggest:

* Making the list of supported dtypes a constant/fixture in the test suite to avoid repetition (forgotten updates).
* Adding a test that (de)serialising the type works when it goes through type inference. What about adding another test that checks `op.add(a, a)` instead of `op.cast(a, to=np.float64)`? Partially done by `test_build_mldtypes` (which curiously checks all dtypes, not just `ml_dtypes`: probably better named `test_dtypes`), but only on the 'input'/serialising side, and not the 'output'/deserialising side. 

I think I'd be most worried about anything being hardcoded for dtypes to be numpy-builtins-only, but I'd assume there's only some `numpy.dtype` subclass checks which should be fine?

PR #266, opened 2025-12-28
Allow inline without valueprop

cbourjau:
Calling `inline` while the value propagation is disabled currently causes a `RuntimeException`. This PR fixes that. 

This is a rather uncommon setup, but came up while working on https://github.com/cbourjau/onnx-tests

# Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@
 Change log
 ==========
 
+0.16.1 (unreleased)
+-------------------
+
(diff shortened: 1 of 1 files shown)


cbourjau:
Thanks a lot for the feedback! You are 100% right, this was a bit hacky! The entire code around `get_backend_calls` is also not soundly typed. I took this as an opportunity to thoroughly refactor the value propagation code. I'll open a new PR.

PR #267, opened 2025-12-30
Refactor value propagation and fix inlining with `_VALUE_PROP_BACKEND.NONE`

cbourjau:
Value propagation currently has a few issues:
- The code around `get_backend_calls` is not soundly typed
- Many of the safeguards and workarounds were written when the reference implementation was in a much worse state.
- `inline` fails when value propagation is disabled.

This PR refactors the value propagation code to address these issues. This supersedes #266. I also tested these changes against the ndonnx test suite using the reference implementation. I recommend starting the review by looking at `_value_prop.py` in its new state rather than the diff.

# Checklist


cbourjau:
> As the implementation is now cleaner, is there any outlook to upstreaming something like #202 or was any other way to use other back-ends introduced in the meantime?

We should indeed finally provide a public way to do that! We currently rely on some private APIs to load custom operators in a proprietary project. It's just perpetually on a low priority. I think this may change when we soon have a second project that requires the same.

PR #313, opened 2026-07-28
Update lockfile and opset definitions

cbourjau:
Update lock file, rerender existing opsets using the latest ONNX version and add operator defintions form `"ai.onnx"` version 25, 26, and 27.

I introduced an upper pin on ruff since the latest version added a lot of new lints. I'll fix those in a separate PR.

# Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,7 +39,7 @@ repos:
       # mypy
       - id: mypy
         name: mypy
-        entry: pixi run -e default mypy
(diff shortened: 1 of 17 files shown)


jbachurski (approved):
Might be worth adding tests if there are any new interesting operators.

cbourjau:
Thanks for the review, @jbachurski ! The new operators are not special from Spox perspective as far as I'm aware 🤞 .

PR #314, opened 2026-07-28
Fix split constructor function

### LINKED ISSUE #304: `op.split`'s `split` input is unusable
The wrapper derives the output count from `num_outputs` only (`out_variadic=num_outputs`) and emits that same value as the node attribute. So with a `split` input there's no valid call:

```python
x = argument(Tensor(np.float32, (6,)))
split = op.const(np.array([1, 2, 3], dtype=np.int64))

op.split(x, split, axis=0)
# Mismatch between number of splits (3) and outputs (1)   # out_variadic defaulted to 1

op.split(x, split, axis=0, num_outputs=3)
# Both 'split' input and 'num_outputs' attribute were given   # ONNX rejects the pair
```

`num_outputs` is the only knob for output count, but setting it also writes the `num_outputs` attribute, which is mutually exclusive with `split`. The explicit-sizes form of Split is therefore unreachable.

cbourjau:
Closes #304 

# Checklist


jbachurski (approved):
LGTM.
One comment: I wonder if it's the right call to look at the propagated value to compute the number of outputs. I wonder if most runtimes might reject that?

cbourjau:
Thanks again! You are right about the concern with the propagated value. I removed that part now. The number of outputs is determined solely by the static shape. If that is not available, an error is raised. If a user knows the shape of a Var that somehow lost its shape information, it can be easily set using `op.reshape`.

